PR #171, opened 2024-10-22
fix: only enable log plugin in production

mattyg:
Avoid inconsistent issue where app crashes due to log plugin being initialized twice on app un-pause. This is a temporary workaround to an upstream issue -- should investigate and report upstream first.

**todo**
- [ ] review and confirm this resolves behavior
- [ ] open upstream issue / pr in tauri-plugin-log and track here


mattyg:
Issue has not appeared again. I believe tauri resolved it.

PR #195, opened 2024-10-27
Release script

mattyg:
- building release packages via CI
- change app icon

**todo**
- [ ] confirm we actually want to change the app icon


mattyg:
cherry-picking the files that are relevant in another PR #196

PR #196, opened 2024-10-27
Ci/automated release builds

mattyg:


mattyg:
Due to the multi-version repo setup, I'm thinking we may want to make this a manual "push-button" trigger, rather than running when a tag is pushed.

PR #197, opened 2024-10-27
feat: update app icons & brand images

mattyg:
Update app icons.

This was extracted from branch `release-script`

**todo**
- [x] ensure updated icons are *ONLY* for the desktop apps, mobile apps have original icon with colon


mattyg:
@zippy can you confirm they still want to switch to these icons?

mattyg:
Eric confirmed: the new icons with the V are for the desktop apps.

The phone app should be the one with the colon.

Issue #54, opened 2024-06-18
Build 2 flavors of app: one with holochain conductor bundled, and the other relying on the Android Service Runtime


tibetsprague:
AC:

- App builds are published for two versions:
  - Light version: 
      - calls into foreground service (intended for use by volla devices)
      - uses "colon" icon
      - Displays notice if unable to connect to conductor, prompting user to start service
  - Rich version: 
      - bundles a standalone holochain conductor, which runs only when the app is open (intended for use by other android devices)
      - uses "squiggle" icon

PR #245, opened 2024-11-17
Feat/holochain foreground service consumer

mattyg:
Partially addresses #54 

- Add 2 feature flags: `holochain_bundled` for the app "rich" version with a bundled conductor and `holochain_service` for the "lite" version that relies on the external holochain service from https://github.com/holochain/mobile-conductor-admin 
- App builds are published for two versions in CI

## todo
- [ ] Rich version of android app uses squiggle icon
- [ ] In the rich version of the app, if it fails to connect to the holochain service, the app displays an error notice with a button to open the settings page for the holochain service settings.
- [ ] release descriptions generated by CI explain the difference between the "rich" and "lite" versions


mattyg:
Superseded by #492

Issue #155, opened 2024-10-16
Camera permission "request every time" not working


mattyg:
The dialog to request permission is not displayed as expected. This is likely an upstream bug with tauri-plugin-barcode-scanner, but tracking here.

wurzer:
Not reproducible any more

PR #289, opened 2024-12-05
selecting messages and option for downloading image

### LINKED ISSUE #115: Make it possible to download images

### LINKED ISSUE #133: Select a message and message actions toolbar
- Long press a message to select it (indicate in UI that it is selected)
- When a message is selected, display actions toolbar
- Potential Actions toolbar actions
   - Copy (copies full content of message)
   - Reply (not implemented yet in DNA)
   - Forward (not implemented yet in DNA)|
   - Hide message
   - Block User
   - Delete (if I authored the message)

MasterK0927:
Resolves #115  Resolves #133

Long press to select message working, we can see Cancel, copy and download on the selected messages in the action bar that appears beneath the selected message, as per the figma design. 

The buttons appearing in the toolbar depends on the type of message, for text download will not appear and for image copy will not appear.

### DIFF
--- a/ui/src/routes/conversations/[id]/+page.svelte
+++ b/ui/src/routes/conversations/[id]/+page.svelte
@@ -15,7 +15,7 @@
   import { t } from '$lib/translations';
   import { copyToClipboard, isMobile, linkify, sanitizeHTML, shareText } from '$lib/utils';
   import { RelayStore } from '$store/RelayStore';
-  import { Privacy, type Conversation, type Message, type Image } from '../../../types';
+  import { Privacy, type Conversation, type Message, type Image, type MessageAction } from '../../../types';
   import LightboxImage from '$lib/LightboxImage.svelte';
 
   // Silly hack to get around issues with typescript in sveltekit-i18n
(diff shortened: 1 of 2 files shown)


ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -80,6 +80,37 @@
...
+      console.error('Image not loaded');
+      return;
+    }
+    if (!image.dataURL) {
+      console.error('Image dataURL is undefined');
+      return;
+    }
+    try {
+      const imageBytes = base64ToBytes(image.dataURL);
+      const blob: Blob = new Blob([imageBytes], { type: 'image/png' });
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:94]: We don't know that the image is a png. Gif and jpg are also supported. We'll want to use the type from `image.type`

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -80,6 +80,37 @@
...
+      const link: HTMLAnchorElement = document.createElement('a');
+      link.href = URL.createObjectURL(blob);
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:96]: The image.dataURL is already a base64 url string, so we shouldn't need to convert it from base64 url string -> blob -> image -> base64 url string. I believe can just use the original base64 url string.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -236,6 +267,115 @@
       })
     }
   }
+
+  const selectedMessages: Writable<Set<string>> = writable(new Set());
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:271]: This doesn't need to be a `writable`. In a svelte component we can just use `let var = 'abc'` and svelte will handle the reactivity for you. For example:

```javascript
let myList = []; 

function addItem(newItem) {
   myList = [...myList, newItem]; 
}
```

Now we can just use `myList` in our component and it will update reactively.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -236,6 +267,115 @@
...
+    
+    function toggleMessageSelection(messageHash: string) {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:273]: Lets only support selecting a single message at a time for now.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -236,6 +267,115 @@
...
+            return message?.content;
+          })
+          .filter(Boolean)
+          .join('\n');
+        copyToClipboard(messagesToCopy);
+      }
+    }, {
+      id: 'download',
+      label: 'Download',
+      icon: 'caretDown',
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:309]: Please use a "download" icon instead of a "caret" icon. You can get svgs of icons from popular icon packs from [iconify](https://icon-sets.iconify.design/?query=download&search-prefix=ic), and add it to the file `lib/svgIcons.ts`. I believe the ones we are using are mostly from the google material icons set, so that would be the first choice if there is one available.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -236,6 +267,115 @@
...
+          const newSelection = new Set(current);
+          Array.from(selectedMessageHashes).forEach(hash => {
+            newSelection.delete(hash);
+          });
+          return newSelection;
+        });
+      }
+    }];
+
+  //Function to handle long press
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:342]: We're using the library [svelte-gestures](https://github.com/Rezi/svelte-gestures) for mobile gesture handling. Please use its included 'on:press` event instead of implementing our own long press logic.

Comment (mattyg) [reply ui/src/routes/conversations/[id]/+page.svelte:342]: Also lets use a single click to select the message on desktop, and only require long-press for mobile.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -236,6 +267,115 @@
...
+          const message = conversation?.data.messages[hash];
+          if (message?.images) {
+            const loadedImages = message.images.filter(img => img.status === 'loaded');
+            await Promise.all(loadedImages.map(downloadImage));
+          }
+        }));
+      }
+    },{
+      id: 'cancel',
+      label: 'Cancel',
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:329]: If you add any strings to the app that are displayed to the user, please render them with the `sveltekit-i18n` library to ensure we can support multiple languages.

See `lib/translations/en/conversations.json` for the english strings we're using. If you don't see a string there you can add it. Other languages will fallback to english if a translation is not found.

Also if you add a new string please note this in the PR description so we remember to notify volla that we need translations for it.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -236,6 +267,115 @@
...
+        } else {
+          newSelection.add(messageHash);
+        }
+        return newSelection;
+      });
+    }
+
+    const messageActions: MessageAction[] = [{
+      id: 'copy',
+      label: 'Copy',
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:287]: Please change the text of the copy button from "Copy" to "Copy Text" to clarify that it does not copy the image.

mattyg (changes requested):
Thanks for getting this up so quickly! I've requested some changes. Please let me know if anything is unclear or you want to go over it together on a call.

- Branch Naming: please use a descriptive name for PR branches. I like to follow the convention of prefix the branch name with `feat/`, `fix/`, `chore/`, `doc/`, `build/` similar to conventional commits (https://www.conventionalcommits.org/en/v1.0.0/). At the very least please make the branch name reflect its content. For example, instead of naming this branch `feature`, it could be named something that describes its content such as `feat/download-images`.

- Code Formatting: please install the recommended vscode extensions from `.vscode/extensions.json`. This will ensure that we're both using the same formatter, and the formatter runs when a file is saved so we have consistent formatting throughout the codebase and don't end up with a bunch of formatting diffs in every PR. Let me know if you have any trouble with the setup in vscode (or you use a different editor).

- PR description: in your PR description, please use the text `Resolves #155 Resolves #133` instead of `this will fix #155 #133`. Including that exact text of "Resolves " will cause github to automatically close the linked issue when the PR is merged.

- Svelte conventions: please go over the [tutorial for svelte v4](https://v4.svelte.dev/) to get comfortable with the best practices / conventions. Svelte provides a lot for you with very simple syntax which makes the code way more readable if you do things its way. So its worth learning and following the "svelte way" of implementing things.

I made a few comments inline, and here are a few more that are easier to list out.

- Image download implementation
	- It doesn't actually seem to download the file for me (on ubuntu 22.04). Lets use [tauri-plugin-dialog](https://v2.tauri.app/plugin/dialog/) for handling file saving cross-platform.
 	
- Message selection
    - Please create a new component MessageActions to encapsulate the logic and rendering of the message action buttons. It can take the Message as an input prop.
    - Clicking outside the selected message should deselect it.
    - The buttons should be styled similarly to other buttons in the app (see the colors used in the button in `routes/account/+page.svelte`)
    - If a message contains both images and text, then we should still have a copy button for copying the text.

ui/src/routes/conversations/[id]/+page.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/+page.svelte:96]: I tried it without the base64tobytes conversion, but it was downloading a corrupted image. 

For viewing the downloaded image, we have to check **Donwload** folder. 

Anyway we have to replace it with plugin.

MasterK0927:
This pull request has been migrated to #292  , please refer to the new pull request for further changes.

Issue #137, opened 2024-10-14
Blank screen when a message arrives to a conversation before the agent profile syncs


tibetsprague:
in ConversationSummary.svelte: 18
$: lastMessageAuthor = $lastMessage ? $conversation.agentProfiles[$lastMessage.authorKey].fields.firstName : null

TypeError: Cannot read properties of undefined (reading 'fields')

tibetsprague:
in ConversationSummary.svelte: 18 
  $: lastMessageAuthor = $lastMessage ? $conversation.agentProfiles[$lastMessage.authorKey].fields.firstName : null

TypeError: Cannot read properties of undefined (reading 'fields')

PR #318, opened 2024-12-16
Adds file sharing functionality and fixes many known bugs

### LINKED ISSUE #124: Add support for file attachments
- supports any file type
- download file
- file icon indicates file type (either with text or different icon)

### LINKED ISSUE #317: Color mismatch for selected message background, copy, and download buttons
When selecting a message, the background, copy, and download button colors are mismatched. While these elements appear as expected in dark mode, switching between modes (e.g., light to dark) causes the colors to persist incorrectly. This results in poor blending and some parts becoming invisible or hard to read.

Introduced in #292 

**Steps to Reproduce**
- Open the app and enable dark mode.
- Select a message and observe the background, copy, and download button colors.
- Switch to light mode and observe the mismatched colors.
- Switch back to dark mode and notice the improper persistence of colors.

**Expected Behavior**
- The selected message's background, copy, and download button colors should adapt appropriately to the theme (light/dark mode), ensuring visibility and proper blending.

MasterK0927:
Resolves #124  
Resolves #317 

1. Allows user to share any kind of file.
2. Currently limits them to 15Mb uploads at max (creating separate issue for it).
3. Thumbnail for pdf and image files, fallback icons for others.
4. Fixes the color mismatch issue reported in #317 
5. Adds an option to cancel or deselect selected files before sending (see image attached for reference). 

Strings added to common.json
`large_file_error`


dnas/relay/zomes/integrity/relay/src/message.rs
```diff
@@ -14,7 +14,7 @@ pub struct File {
 pub struct Message {
     pub content: String,
     pub bucket: u32,
-    pub images: Vec<File>,
+    pub files: Vec<File>,
```
mattyg: Ahh, I didn't realize this is what you were referring to in your question. 

So because this changes the code of an integrity zome, it also changes the DNA hash of the `relay` DNA, and thus is a network breaking-change. This means that users on a prior version of the app will not be in the same network as users on this version.

I agree the naming here is not accurate, so I do want to change it, but we'll need to be careful about how its done.

I suggest the following:
- Modify this PR to be only the *non-breaking* changes necessary to implement this feature (i.e. all 
- This PR gets merged into `develop` and backported to `develop-0.3`
- Open another PR with the *breaking* changes
- The breaking PR gets merged into `develop` -- **but does not get backported to `develop-0.3`**

Thus when we are ready to update the app to 0.4, which is itself a network breaking change, we will also include this breaking change as well.

While I would generally prefer consistent naming in the front-end and back-end, I see that it probably makes sense to make an exception here to avoid the breaking change. But also, lets keep the scope of differences between `develop` and `develop-0.3` as minimal as possible (i.e. lets use the new naming "files" in the front-end, except when making requests to the backend).

package.json
```diff
@@ -49,6 +49,7 @@
     "@tauri-apps/plugin-dialog": "^2.0.1",
     "dompurify": "^3.1.6",
     "lodash-es": "^4.17.21",
+    "pdfjs-dist": "^4.8.69",
```
mattyg, line 52: I think this is a nice idea, but was not a requirement to resolve the issue. 

In the future, I would recommend making an issue for it so it can be prioritized against others. If we are going to support PDF previews, then I would want to support other file type previews as well, with a unified implementation as much as possible.

For now I would suggest removing this functionality from this PR and opening a separate PR with it, so we don't loose track of the work.

ui/src/lib/PDFThumbnail.svelte
mattyg, line 1: I'll avoid commenting on this, and we'll do it in a separate PR with just this PDF thumbnail feature, if that's okay with you.

ui/src/lib/svgIcons.ts
```diff
@@ -62,4 +62,24 @@ export const svgIcons: { [key: string]: string } = {
     '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path fill="evenodd" d="M1 21h22L12 2zm12-3h-2v-2h2zm0-4h-2v-4h2z"/></svg>',
   download:
     '<svg xmlns="http://www.w3.org/2000/svg" width="60" height="60" viewBox="0 0 24 24"><path d="M16.59 9H15V4c0-.55-.45-1-1-1h-4c-.55 0-1 .45-1 1v5H7.41c-.89 0-1.34 1.08-.71 1.71l4.59 4.59c.39.39 1.02.39 1.41 0l4.59-4.59c.63-.63.19-1.71-.7-1.71M5 19c0 .55.45 1 1 1h12c.55 0 1-.45 1-1s-.45-1-1-1H6c-.55 0-1 .45-1 1"/></svg>',
+  file: '<svg width="800" height="800" viewBox="-6 0 46 46" xmlns="http://www.w3.org/2000/svg"><path data-name="12.File" d="M30 46H4a4 4 0 0 1-4-4V4a4 4 0 0 1 4-4h19a.98.98 0 0 1 .776.395l9.829 9.829A.98.98 0 0 1 34 11v31a4 4 0 0 1-4 4M24 3.381V10h6.619ZM32 12h-9a1 1 0 0 1-1-1V2H4a2 2 0 0 0-2 2v38a2 2 0 0 0 2 2h26a2 2 0 0 0 2-2Zm-6 26H8a1 1 0 0 1 0-2h18a1 1 0 0 1 0 2m0-8H8a1 1 0 0 1 0-2h18a1 1 0 0 1 0 2m0-8H8a1 1 0 0 1 0-2h18a1 1 0 0 1 0 2" fill-rule="evenodd"/></svg>',
+  audio:
```
mattyg, line 66: This is specifically for `mp3` not audio

ui/src/lib/svgIcons.ts
```diff
@@ -62,4 +62,24 @@ export const svgIcons: { [key: string]: string } = {
...
+  video:
```
mattyg, line 68: This is specifically for mp4 files, not all videos.

ui/src/lib/svgIcons.ts
mattyg, line 1: We should have consistent styling of all file icons, rather than taking icons with specific file extensions from different icon packs. To accomplish this I think we have 2 different options:
- Option 1: We support a couple very-common file type icons, while all others fall back to a generic file icon. The common file type icons are *not* specific to a file extension, but rather a type of file. I.e. we have icons for document, audio, video, and a fallback generic file icon.
- Option 2: We create a component to render a generic file icon, with any arbitrary file extension text inside of it. Then when rendering the component we specify the file extension text.

Either option is okay with me, I don't feel strongly. You could ask for feedback if you don't either.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -212,31 +214,40 @@
...
+
+      const readers: Promise<FileMetadata>[] = validFiles.map((file) => {
         const reader = new FileReader();
         reader.readAsDataURL(file);
-        return new Promise<Image>((resolve) => {
+        return new Promise<FileMetadata>((resolve) => {
           reader.onload = async () => {
             if (typeof reader.result === "string") {
               resolve({
+                id: crypto.randomUUID(),
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:247]: I don't understand the utility of this.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -316,6 +332,78 @@
       }
     }
   }
+
+  function formatFileSize(size: number): string {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]: Please find a well-maintained js library to provide this utility for us.

Here's one option: https://www.npmjs.com/package/pretty-bytes

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -316,6 +332,78 @@
...
+    const units = ["B", "KB", "MB", "GB", "TB", "PB"];
+    let unitIndex = 0;
+    while (size >= 1024 && unitIndex < units.length - 1) {
+      size /= 1024;
+      unitIndex++;
+    }
+    return `${size.toFixed(2)} ${units[unitIndex]}`;
+  }
+
+  function formatFileName(file: FileMetadata, maxCharacters: number = 10): string {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]: I don't understand the purpose of this. Why does the file name need to be altered?

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -316,6 +332,78 @@
...
+          return parts[0].slice(0, maxCharacters) + "..." + extension;
+        }
+      }
+      return baseName.slice(0, maxCharacters) + "..." + extension;
+    }
+    return fileName;
+  }
+
+  //To get the icon name based on the file type
+  function formatFileIcon(file: FileMetadata): string {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:349]: This will be altered by the changes to icons from the previous comment.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -521,13 +609,13 @@
               {/if}
               <li
                 class="mt-auto {!message.hideDetails && 'mt-3'} relative {isSelected
-                  ? 'bg-secondary-500 mb-20 mt-2 rounded-t-xl'
-                  : ''}"
+                  ? 'bg-tertiary-500 dark:bg-secondary-500 mb-20 mt-2 rounded-t-xl'
+                  : ''} "
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]:
```suggestion
                  : ''}"
```

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -557,21 +645,71 @@
...
-                              />
+                    <!-- if message contains files -->
+                    {#if message.files && message.files.length > 0}
+                      {#each message.files as file}
+                        <div class="flex {fromMe ? 'justify-end' : 'justify-start'} w-full p-2">
+                          <!-- if file is loaded -->
+                          {#if file.status === "loaded"}
+                            <div class="mb-2 flex w-full max-w-full items-start justify-between">
+                              <!-- Display image thumbnail -->
+                              {#if file.fileType.startsWith("image/")}
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:620]: :+1:

ui/src/routes/conversations/[id]/+page.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/+page.svelte]: Please see the reference image, we can show a limited text in the text area. So if file name is too big, we can truncate it.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -557,21 +645,71 @@
...
+                                <div class="w-full">
+                                  <LightboxImage
+                                    btnClass="inline w-full sm:max-w-md lg:max-w-lg transition-all duration-200"
+                                    src={file.dataURL}
+                                    alt={file.name}
+                                  />
+                                </div>
+                                <!-- Display pdf thumbnail -->
+                              {:else if file.fileType === "application/pdf"}
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:665]: This will be pulled out temporarily to review in a separate PR

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -640,21 +771,104 @@
...
+                <img src={file.dataURL} class="h-16 w-16 rounded-lg object-cover" alt="thumbnail" />
+                <button
+                  class="absolute -right-2 -top-2 flex h-5 w-5 items-center justify-center rounded-full bg-white p-1 text-white"
+                  on:click={() => cancelUpload(file.id)}
+                  aria-label="Cancel Upload"
+                >
+                  <SvgIcon icon="x" size="8" />
+                </button>
+              </div>
+            {:else if file.dataURL && file.fileType.startsWith("application/pdf")}
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:809]: This will be pulled out temporarily to review in a separate PR, if thats okay

ui/src/routes/conversations/[id]/+page.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/+page.svelte:247]: It's used for deselecting the files we want to send.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -316,6 +332,78 @@
...
+    };
+    if (file.name) {
+      const extension = file.name.split(".").pop()?.toLowerCase();
+      if (extension && extensionIcons[extension]) {
+        console.log(extensionIcons[extension]);
+        return extensionIcons[extension];
+      }
+    }
+    //Final fallback
+    return "file";
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:369]: :+1:

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -316,6 +332,78 @@
...
+      mkv: "mkv",
+    };
+    if (file.name) {
+      const extension = file.name.split(".").pop()?.toLowerCase();
+      if (extension && extensionIcons[extension]) {
+        console.log(extensionIcons[extension]);
+        return extensionIcons[extension];
+      }
+    }
+    //Final fallback
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]:
```suggestion
    // Final fallback
```

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -316,6 +332,78 @@
...
+        }
+      }
+      return baseName.slice(0, maxCharacters) + "..." + extension;
+    }
+    return fileName;
+  }
+
+  //To get the icon name based on the file type
+  function formatFileIcon(file: FileMetadata): string {
+    //Extension based fallback
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]:
```suggestion
    // Extension based fallback
```

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -316,6 +332,78 @@
...
+          }
+          return parts[0].slice(0, maxCharacters) + "..." + extension;
+        }
+      }
+      return baseName.slice(0, maxCharacters) + "..." + extension;
+    }
+    return fileName;
+  }
+
+  //To get the icon name based on the file type
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]:
```suggestion
  // To get the icon name based on the file type
```

package.json
MasterK0927: Actually Dr Wurzer wanted to have thumbnail for pdfs, so I added it.

ui/src/lib/translations/en/common.json
```diff
@@ -41,5 +41,6 @@
   "create_conversation_error": "Failed to create conversation",
   "add_contact_to_conversation_error": "Failed to add contact to conversation",
   "download_file_error": "Failed to download file",
-  "download_file_success": "File downloaded successfully"
+  "download_file_success": "File downloaded successfully",
+  "large_file_error": "Currently we don't support files larger than {{maxSize}}"
```
mattyg:
```suggestion
  "large_file_error": "Attached files are limited to {{maxSize}}"
```

ui/src/routes/conversations/[id]/+page.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/+page.svelte]: I would suggest we simplify this significantly:

- Option 1: slice the string to a maximum characters, then append "..." 
- Option 2: render the full string, but use the css property `text-overflow: ellipsis;`

Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/+page.svelte]: Current impl allows us to specify the maxLength we want, so that we can use it in, multiple usecases. And it aligns with the first option.

mattyg (changes requested):
Very nice! Made some comments inline.

- Can you please make a ticket "Translations Needed" with a checklist of all the translations we need from the recent 3 PRs? 

- Files displayed in chat should align to the right of the page, and not be streached to the length of images below
image

- File type icons are different sizes (see the screenshot above)

- If I attach a file, then remove it, then attach the same file again, it is not added to the message.

- If I attach a file, then remove it, then click the attach button again, the previous file is selected by default. I think no files should be selected.

package.json
mattyg: Ah okay. Then we'll keep it in this PR and I'll update my comments.

ui/src/lib/PDFThumbnail.svelte
```diff
@@ -0,0 +1,51 @@
...
+  export let width: number;
+  export let height: number;
+  export let fallbackIcon: string = "file";
+
+  let thumbnailUrl: string | null = null;
+
+  onMount(async () => {
+    try {
+      const pdfjsLib = await import("pdfjs-dist");
+      pdfjsLib.GlobalWorkerOptions.workerSrc = `//cdnjs.cloudflare.com/ajax/libs/pdf.js/4.8.69/pdf.worker.min.mjs`;
```
mattyg: Can you explain:
- why does import happen when mounting the component?
- why are we pulling from a CDN rather than bundling the dependency? Can we include it in our normal package.json dependencies?

ui/src/routes/conversations/[id]/+page.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/+page.svelte:247]: We should be able to just use the index of the file in the list, right?

ui/src/routes/conversations/[id]/+page.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/+page.svelte]: I think the implementation is overly complex which will make it harder to read and maintain. I also don't think it adds value to include the file extension in the string which can make it simpler. Please distill it down to just what I outlined above.

package.json
mattyg: Next time, if there are additional requirements, please make an issue for them or add them to the existing issue.

ui/src/lib/PDFThumbnail.svelte
```diff
@@ -0,0 +1,51 @@
...
+
+{#if thumbnailUrl}
+  <img
+    src={thumbnailUrl}
+    alt="PDF Thumbnail"
+    class="rounded object-cover"
+    style="width: {width}px; height: {height}px;"
+  />
+{:else}
+  <SvgIcon icon={fallbackIcon} color={$modeCurrent ? "black" : "white"} size={width.toString()} />
```
mattyg: :+1:

mattyg:
Additional comments related to PDF generation:

- When I attach a PDF, I do not see a preview, and get the following error in console (on ubuntu 22.04)
image. It does fall back to the pdf icon which is great.

mattyg:
Also please update the description to say:

Resolves #124
Resolves #137 

(currently only the first ticket will be automatically closed when the PR is merged. github isn't that smart :) )

ui/src/routes/conversations/[id]/+page.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/+page.svelte:247]: I tried with index first, but with the similar files there was issues, so I chose id as the parameter.

ui/src/lib/PDFThumbnail.svelte
MasterK0927: - Import is happening during component mounting for achieving lazy loading, to reduce the initial bundle size by loading `pdfjs-dist` only when the component is needed.

- I’ve faced import-related issues with `pdfjs-dist` in previous projects as well. While using it with React/Next.js, I encountered the same problem, and it wasn’t resolved for me either (refer to [this](https://github.com/vercel/next.js/issues/58313) thread). I ended up migrating to `react-pdf`. There’s something similar to it, like `svelte-pdf`, but it’s not maintained. So I switched to cdn approach.

MasterK0927:
Made all then changes according to the comments.

image
image

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -333,75 +337,35 @@
...
-      mp3: "audio",
-      mp4: "video",
-      mkv: "mkv",
+    const commonFileTypes: { [key: string]: string[] } = {
+      document: ["doc", "docx", "rtf", "txt", "odt"],
+      spreadsheet: ["xls", "xlsx", "csv", "ods"],
+      presentation: ["ppt", "pptx", "odp"],
+      audio: ["mp3", "wav", "ogg", "flac"],
+      video: ["mp4", "avi", "mkv", "mov"],
+      archivefile: ["zip", "rar", "7z", "tar", "gz"],
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:356]: Just a note that it might be simpler to match based on mime type rather than file extension (i.e. mime types for video are generally `video/*`, audio is `audio/*`, text is `text/*`, etc.). But this is fine.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -333,75 +337,35 @@
...
-          }
-          return parts[0].slice(0, maxCharacters) + "..." + extension;
-        }
-      }
-      return baseName.slice(0, maxCharacters) + "..." + extension;
+    if (fileName.length <= maxLength) {
+      return fileName;
     }
-    return fileName;
+    return fileName.slice(0, maxLength) + "...";
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte:345]: :+1:

ui/src/lib/PDFThumbnail.svelte
mattyg: > Import is happening during component mounting for achieving lazy loading, to reduce the initial bundle size by loading pdfjs-dist only when the component is needed.

I wouldn't worry about initial bundle size. Since this is a desktop/mobile app, the user already has the initial bundle locally and just renders it, so its not the big performance hit like on the web where the user has to fetch the initial bundle before they can render anything.

> I’ve faced import-related issues with pdfjs-dist in previous projects as well. While using it with React/Next.js, I encountered the same problem, and it wasn’t resolved for me either (refer to https://github.com/vercel/next.js/issues/58313 thread). I ended up migrating to react-pdf. There’s something similar to it, like svelte-pdf, but it’s not maintained. So I switched to cdn approach.

I dug into some github issue threads on the pdfjs repo and was able to get it working with the local worker included in the the npm package. I pushed the change to this branch. Please take a look at how it was implemented.

mattyg (changes requested):
Looking good - almost there!

- The fallback pdf icon is larger than the others
image
- The fallback file icon looks like a different design than the others (it has a triangle on the top right corner). Can we get the designs to be consistent?
- When I attach a file, then press the enter key, the file is removed from the message.
- There's a bunch of merge conflicts that need to be resolved before we can merge this
- Can you please make a github issue for supporting downloading only a single file (when a message has multiple files)? I'm guessing we might want this sooner rather than later.

mattyg:
> The fallback pdf icon is larger than the others

This has not been addressed. See the screenshot below. The pdf fallback icon is noticeably larger than the other file type icons.

> There's a bunch of merge conflicts that need to be resolved before we can merge this

This still needs to be addressed before I can merge.

mattyg (changes requested):
> The fallback pdf icon is larger than the others

This still needs to be addressed. See the screenshot below -- the pdf fallback icon is noticeably larger than the other file type icons.
image


> There's a bunch of merge conflicts that need to be resolved before we can merge this

This still needs to be addressed before I can merge.

MasterK0927:
> > The fallback pdf icon is larger than the others
> 
> This still needs to be addressed. See the screenshot below -- the pdf fallback icon is noticeably larger than the other file type icons. image
> 
> > There's a bunch of merge conflicts that need to be resolved before we can merge this
> 
> This still needs to be addressed before I can merge.

I reviewed the merge conflicts, and they involve breaking changes. Some files have been completely modified, while certain parts have been migrated to new components. What's the best way to handle them?

MasterK0927:
> > The fallback pdf icon is larger than the others
> 
> This still needs to be addressed. See the screenshot below -- the pdf fallback icon is noticeably larger than the other file type icons. image
> 
> > There's a bunch of merge conflicts that need to be resolved before we can merge this
> 
> This still needs to be addressed before I can merge.

@mattyg For me its working fine, can you please check it once again doing a git pull ?
Screenshot from 2024-12-24 13-29-41

mattyg (changes requested):
> @mattyg For me its working fine, can you please check it once again doing a git pull ?
- Whoops my bad, it looks good to me too. Sorry about that.

Sorry just another 2 minor changes I just noticed:
- In the message input, the pdf icon is a different height (it looks like the pdf icon is the proper height, but the other icons should be taller)
image

- When I receive messages with files from other users (i.e. when the message is displayed on on the left side of the page), the file icon should be on the left side of the file title & size.
image


- Sorry there are so many merge conflicts. I've done a lot of refactoring over the past week and will be doing more this week. I'm guessing it might be easiest to make a new branch off develop and copy the relevant changes into it.

MasterK0927:
This has been moved to #369

PR #369, opened 2024-12-26
feat: file sharing and upload cancelling

### LINKED ISSUE #124: Add support for file attachments

### LINKED ISSUE #317: Color mismatch for selected message background, copy, and download buttons

MasterK0927:
Synced version of #318 

Resolves https://github.com/holochain-apps/volla-messages/issues/124
Resolves https://github.com/holochain-apps/volla-messages/issues/317

- Allows user to share any kind of file.
- Currently limits them to 15Mb uploads at max (creating separate issue for it).
- Thumbnail for pdf and image files, fallback icons for others.
- Fixes the color mismatch issue reported in https://github.com/holochain-apps/volla-messages/issues/317
- Adds an option to cancel or deselect selected files before sending.
- Strings added to common.json
    `large_file_error`


package.json
```diff
@@ -45,5 +47,9 @@
     "tslib": "^2.6.2",
     "typescript": "^5.4.4"
   },
-  "dependencies": {}
+  "dependencies": {
+    "@svelte-put/clickoutside": "^4.0.0",
+    "pdfjs-dist": "^4.9.155",
+    "pretty-bytes": "^6.1.1"
+  }
```
mattyg: None of these belong in the root package.json. They are all used only in the `ui` workspace, so they belong in the `ui/package.json`

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -222,6 +221,38 @@
...
+        for (const [type, extensions] of Object.entries(commonFileTypes)) {
+          if (extensions.includes(extension)) {
+            return type;
+          }
+        }
+      }
+    }
+    // Final fallback
+    return "file";
+  }
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]: For clearer organization, please extract these functions into a standalone component `FileIcon` which will handle rendering the icon for a given File.

Then the `FileIcon` component can be used in both `Message` and `ConversationMessagesInput`

ui/src/routes/conversations/[id]/ConversationMessageInput.svelte
```diff
@@ -12,21 +15,33 @@
     };
   }>();
 
+  const MAX_FILE_SIZE = 15 * 1024 * 1024; // 15MB
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessageInput.svelte]: Lets keep all configuration together in `ui/src/config.ts` for ease of configuration and visibility of which variables are used for app config.

ui/src/translations/en/common.json
```diff
@@ -41,5 +41,6 @@
   "create_conversation_error": "Failed to create conversation",
   "add_contact_to_conversation_error": "Failed to add contact to conversation",
   "download_file_error": "Failed to download file",
-  "download_file_success": "File downloaded successfully"
+  "download_file_success": "File downloaded successfully",
+  "large_file_error": "Attached files are limited to {{maxSize}}"
```
mattyg: These messages would only be rendered in the `/conversations` routes, right? If so, they belong in the `translations/en/conversations.json`

ui/vite.config.ts
```diff
@@ -19,4 +19,10 @@ export default defineConfig({
   define: {
     "window.__APP_VERSION__": JSON.stringify(version), // Define a global constant
   },
+  optimizeDeps: {
+    include: ["pdfjs-dist/legacy/build/pdf.js", "pdfjs-dist/legacy/build/pdf.worker.js"],
+  },
+  ssr: {
+    noExternal: ["pdfjs-dist"],
+  },
```
mattyg: I don't think these custom vite config is necessary, unless I'm missing something.

mattyg (changes requested):
- I requested some changes in comments inline
- I'm not able to run `npm i`, there appear to be conflicting dependencies). I'm not able to review the PR further due to this issue)

ui/src/routes/conversations/[id]/Message.svelte
```diff
@@ -15,6 +15,7 @@
   import PdfThumbnail from "$lib/PdfThumbnail.svelte";
   import { isMobile } from "$lib/utils";
   import prettyBytes from "pretty-bytes";
+  import FileIcon from "../../../lib/FileIcon.svelte";
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/Message.svelte]:
```suggestion
  import FileIcon from "$lib/FileIcon.svelte";
```

ui/src/lib/FileIcon.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
+  import { modeCurrent } from "@skeletonlabs/skeleton";
+  import SvgIcon from "$lib/SvgIcon.svelte";
+  import type { Image } from "../types";
```
mattyg:
```suggestion
  import type { Image } from "$lib/types";
```

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -311,13 +318,14 @@
         <ConversationEmpty {conversation} />
       {:else}
         <!-- Display conversation messages -->
-        <ConversationMessages messages={$processedMessages} />
+        <ConversationMessages messages={$processedMessages} {formatFileName} />
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]: I think we should be able to push this `formatFileName` function down into the `FileIcon` component as well (or if we want to keep the file icon separate, maybe a `FileIconNamed` component that includes it). 

Currently we're duplicating the rendering of file name, icon, and bytes 4 times (2x in `Message`, and 2x in `ConversationMessageInput`), and we have to pass this function down multiple layers. That's a good indicator that it should be extracted into a standalone component.

If there are slight variations in rendering depending on the context, those can be passed to the component as props.

mattyg (changes requested):
This looks great! Just a few minor changes requested in comments.

Sorry I got some other PRs in yesterday so there are merge conflicts again. I'll be doing more today as well.

ui/src/routes/conversations/[id]/+page.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/+page.svelte]: I was thinking for the same.

MasterK0927:
> This looks great! Just a few minor changes requested in comments.
> 
> Sorry I got some other PRs in yesterday so there are merge conflicts again. I'll be doing more today as well.

Then in that case I will wait for that to be done first. Till then I am working on Voice Calls.

MasterK0927:
- Resolves all the merge conflicts
- Migrates the thumbnail previews into a single component

ui/src/lib/Button.svelte
```diff
@@ -20,7 +21,7 @@
     {#if loading}
       <SvgIcon icon="spinner" size={iconSize} color="%23FD3524" />
     {:else if icon !== undefined}
-      <SvgIcon {icon} size={iconSize} color="%23FD3524" />
+      <SvgIcon {icon} size={iconSize} color={$modeCurrent ? "white" : "%23FD3524"} />
```
mattyg: As far as I can tell, the standard button should have a red icon in both light and dark modes -- that's how the figma mockups look. I'm assuming that's the intended styling and anything else we see in the code base is either a non-standard button, or is a mistake in the implementation. Lets revert this and keep the standard buttons with red icons.

ui/src/lib/FileIcon.svelte
mattyg, line 1: :+1:

ui/src/lib/FilePreview.svelte
```diff
@@ -0,0 +1,165 @@
...
+  import FileIcon from "$lib/FileIcon.svelte";
+  import { isMobile } from "$lib/utils";
+  import LightboxImage from "$lib/LightboxImage.svelte";
+
+  export let file: Image;
+  export let showCancel: boolean = false;
+  export let fromMe: boolean = false;
+  export let isMessage: boolean = false;
+  export let maxFilenameLength: number = 10;
+  export let onCancel: ((id: string) => void) | undefined = undefined;
```
mattyg: Instead of passing functions from parent to child, please dispatch events from child to parent. This makes the flow of relationship between parent and child more clear and explicit. 

To accomplish this here you can do something like the following:

```
import { createEventDispatcher } from "svelte";
const dispatch = createEventDispatcher<{ cancel: string }>();
```

```
on:click={() => dispatch("cancel", file.id)
```


Then in the parent:
```
<FilePreview on:cancel={(e) => cancelUpload(e.detail)} />
```

ui/src/lib/FilePreview.svelte
```diff
@@ -0,0 +1,165 @@
...
+  import { modeCurrent } from "@skeletonlabs/skeleton";
+  import prettyBytes from "pretty-bytes";
+  import PdfThumbnail from "$lib/PdfThumbnail.svelte";
+  import FileIcon from "$lib/FileIcon.svelte";
+  import { isMobile } from "$lib/utils";
+  import LightboxImage from "$lib/LightboxImage.svelte";
+
+  export let file: Image;
+  export let showCancel: boolean = false;
+  export let fromMe: boolean = false;
```
mattyg: Rather than having props be specific to the context they are rendered within, lets make them as generic as possible so the component is as generally useful as possible, including in different contexts.

So in this case, lets change `fromMe: boolean = false` to `align: "left" | "right" = "left".

Or actually, even better lets avoid the magic strings by creating a an typescript enum in the `types.ts` file for `Alignment { Left, Right }` and then make the prop `align: Alignment = Alignment.Right`

ui/src/lib/FilePreview.svelte
```diff
@@ -0,0 +1,165 @@
...
+  import SvgIcon from "$lib/SvgIcon.svelte";
+  import { modeCurrent } from "@skeletonlabs/skeleton";
+  import prettyBytes from "pretty-bytes";
+  import PdfThumbnail from "$lib/PdfThumbnail.svelte";
+  import FileIcon from "$lib/FileIcon.svelte";
+  import { isMobile } from "$lib/utils";
+  import LightboxImage from "$lib/LightboxImage.svelte";
+
+  export let file: Image;
+  export let showCancel: boolean = false;
```
mattyg: :+1:

ui/src/lib/Button.svelte
MasterK0927: The button's background color is red by default, and the icon also turns red, causing it to blend in with the background.

So I did this

ui/src/lib/FilePreview.svelte
```diff
@@ -0,0 +1,165 @@
...
+      </div>
+      <div class="flex items-center justify-center">
+        <PdfThumbnail pdfDataUrl={file.dataURL ?? ""} width={30} height={43} fallbackIcon="pdf" />
+      </div>
+    </div>
+  {:else}
+    <div
+      class="bg-surface-800/10 relative mb-2 mr-2 flex flex-row items-start justify-between gap-1.5 rounded-xl p-2"
+    >
+      {#if showCancel}
```
mattyg: We should only need a single `{#if showCancel}` conditional that works for all cases.

ui/src/lib/FilePreview.svelte
```diff
@@ -0,0 +1,165 @@
...
+  function formatFileName(file: Image, maxLength: number = maxFilenameLength): string {
+    const fileName = file.name.trim();
+    if (fileName.length <= maxLength) {
+      return fileName;
+    }
+    return fileName.slice(0, maxLength) + "...";
+  }
+</script>
+
+{#if isMessage}
```
mattyg: Lets avoid having props be specific to the parent's context. It seems like the `isMessage` prop is bundling several changes together:
- There is slightly different styling of the background
- There is different sizing of the icons and previews

We should be able to split these into multiple props that make the component more generic for use in different contexts. We may also potentially want to move the  background div styling up to the parent component if that makes it easier.

ui/src/lib/FilePreview.svelte
```diff
@@ -0,0 +1,165 @@
...
+  {#if file.status === "loading" || file.status === "pending"}
```
mattyg: I see that the "pending" status is preventing us from rendering file previews in the the input field. I suggest we add another file status "preview" for this case.

mattyg (changes requested):
Very good! Looking much better!

I requested some changes inline for making the `FilePreview` component more generic and avoid duplicating code. I pushed a few of the changes already and left the rest to you. Let me know if you have questions.

ui/src/lib/FilePreview.svelte
mattyg: (I made this change a pushed it - please take a look so its clear to you :))

MasterK0927: I actually tried it, but was unable to get it working, so I chose to move with props. Thanks, it will clear my confusion.

ui/src/lib/Button.svelte
mattyg: Ah it seems like we should change the default background of the Button component to match the styling used through the app, and remove the additional css classes used to override it in most of the places where it is used.

ui/src/lib/FilePreview.svelte
mattyg: (I made this change a pushed it - please take a look so its clear to you :))

ui/src/lib/Button.svelte
mattyg: For this PR lets leave it as is. I'm currently working through refactoring the button styling.

MasterK0927:
Sorry to override your alignment related changes, but I think this is more generic to our use case.

ui/src/lib/FilePreview.svelte
```diff
@@ -0,0 +1,124 @@
...
+    return fileName.length <= maxFilenameLength
+      ? fileName
+      : `${fileName.slice(0, maxFilenameLength)}...`;
+  }
+
+  $: config = sizeConfig[size];
+  $: isImage = file.fileType.startsWith("image/");
+  $: isPdf = file.fileType === "application/pdf";
+  $: isLoading = file.status === FileStatus.Loading || file.status === FileStatus.Pending;
+  $: hasError = file.status === FileStatus.Error;
```
mattyg, line 49: great!

ui/src/lib/types.ts
```diff
@@ -186,3 +195,8 @@ export interface ContactRecord {
   signed_action: SignedActionHashed;
   contact?: Contact;
 }
+
+export enum Alignment {
```
mattyg, line 218: If you're not using this type, please remove it.

mattyg: I modified the implementation to use this type, to avoid "magic strings"

MasterK0927: Changing it to use Alignment enum was not working for me with my changes, please test it once.

mattyg (approved):
Nice! The FilePreview component looks great! So much simpler and clearer to quickly understand! Thanks for sticking with it.

I made a few minor changes and pushed them:
- using enum types for Alignment and Size, rather than magic string
- making a separate prop to enable the lightbox for images
- using css sflex-row and flex-row-reverse for alignment

I also pulled in the latest develop and resolved the merge conflicts. Next time please check if there are merge conflicts before requesting review :). If you see this message at the bottom of the PR page, then you know there are conflicts:
image

MasterK0927:
> Nice! The FilePreview component looks great! So much simpler and clearer to quickly understand! Thanks for sticking with it.
> 
> I made a few minor changes and pushed them:
> - using enum types for Alignment and Size, rather than magic string
> - making a separate prop to enable the lightbox for images
> - using css sflex-row and flex-row-reverse for alignment
> 
> I also pulled in the latest develop and resolved the merge conflicts. Next time please check if there are merge conflicts before requesting review :). If you see this message at the bottom of the PR page, then you know there are conflicts:
> image

Ahh I doubt I missed on your early morning commits. So merge conflicts emerged.

Issue #334, opened 2024-12-18
Translation Required


MasterK0927:
The following strings need to be translated

**[common](https://github.com/holochain-apps/volla-messages/blob/develop/ui/src/translations/locales/en.json)**

- [ ] copy_text
- [ ] download
- [ ] error_sending_message
- [ ] error_loading_file
- [ ] copy_success
- [ ] copy_error
- [ ] create_account_error
- [ ] share_code_error
- [ ] camera_permission_error
- [ ] scan_error
- [ ] update_profile_error
- [ ] upload_image_error
- [ ] create_conversation_error
- [ ] add_contact_to_conversation_error
- [ ] large_file_error
- [ ] stores_setup
- [ ] stores_setup_error

PR #419, opened 2025-01-07
Refactor/translations one file

mattyg:
- Single file for all translations in a single locale. All are now under the module "common".
- Order translation strings alphabetically by key for ease of identifying missing & duplicates
- Remove several unused translation strings
- Remove some duplicate translation strings
- Rename some translation string keys for clarity


mattyg:
updated #334 to reflect the changes

Issue #405, opened 2025-01-01
Refactor icons to use iconify


mattyg:
Currently we have many icons copied from different icons packs, with different sizes and color defaults. They are included as svgs in the file `svgIcons.ts`.

We should refactor our icons to use https://iconify.design/docs/icon-components/svelte/ to ensure a standardized coloring, sizing, and ideally icon pack.

PR #420, opened 2025-01-08
Implements contact deletion functionality

### LINKED ISSUE #13: Remove a contact
AC:
- I can remove a contact from my contacts list

MasterK0927:
Resolves #13

- Adds the functionality for removing the contact from the contact list.

- Adds a button on the create page (where contacts are listed) to allow users to remove contacts from the list.


MasterK0927:
fill="none" is needed in delete icon for retaining its transparent background behr

ui/src/lib/ButtonDelete.svelte
```diff
@@ -0,0 +1,22 @@
...
+    // It ensures that the click doesn't bubble up to parent elements
+    // It's important when the button is nested inside other clickable elements
+    event.stopPropagation();
+    onDelete(event);
+  }
+</script>
+
+<button
+  type="button"
+  on:click={handleClick}
```
mattyg:
```suggestion
  on:click|stopPropagation
```

Then remove the 'onDelete' prop and handle it as an on:click event from the parent component. No passing functions as props, please.

ui/src/lib/ContactListItem.svelte
```diff
@@ -18,32 +19,45 @@
   onMount(async () => {
     hasAgentJoinedDht = await contactStore.getHasAgentJoinedDht(agentPubKeyB64);
   });
+
+  async function handleDelete(event: MouseEvent) {
+    try {
+      await contact.delete();
+    } catch (err) {
+      console.error(err);
```
mattyg: Need an error toast so something is shown to the user

ui/src/lib/ContactListItem.svelte
```diff
@@ -6,6 +6,7 @@
   import type { AgentPubKeyB64 } from "@holochain/client";
   import Avatar from "$lib/Avatar.svelte";
   import { getContext, onMount } from "svelte";
+  import ButtonDelete from "./ButtonDelete.svelte";
```
mattyg:
```suggestion
  import ButtonDelete from "$lib/ButtonDelete.svelte";
```

ui/src/lib/svgIcons.ts
```diff
@@ -90,4 +90,5 @@ export const svgIcons: { [key: string]: string } = {
...
+  delete: '<svg width="100%" height="100%" viewBox="0 0 25 25" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M5 6.66671H20.8333M19.1667 7.50004V19.1667C19.1667 19.8297 18.9033 20.4656 18.4344 20.9345C17.9656 21.4033 17.3297 21.6667 16.6667 21.6667H9.16667C8.50363 21.6667 7.86774 21.4033 7.3989 20.9345C6.93006 20.4656 6.66667 19.8297 6.66667 19.1667V7.50004M10.8333 6.66671V4.16671C10.8333 3.94569 10.9211 3.73373 11.0774 3.57745C11.2337 3.42117 11.4457 3.33337 11.6667 3.33337H14.1667C14.3877 3.33337 14.5996 3.42117 14.7559 3.57745C14.9122 3.73373 15 3.94569 15 4.16671V6.66671M10.8333 10.8334V16.6667M15 10.8334V16.6667" stroke="#FB0707" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></svg>',
```
mattyg: `fill="currentColor"` doesn't work? I want a consistent solution for all svgs, per the documentation at the top of the file.

ui/src/store/ContactStore.ts
```diff
@@ -80,6 +81,24 @@ export function createContactStore(client: RelayClient): ContactStore {
     );
   }
 
+  /**
+   * Delete a contact
+   */
+  async function deleteContact(agentPubKeyB64: AgentPubKeyB64) {
+    const contact = contacts.getKeyValue(agentPubKeyB64);
+    if (!contact) {
+      throw new Error(`Contact not found for agent ${agentPubKeyB64}`);
```
mattyg: Not necessary -- getKeyValue handles this -- we can trust the return types to be accurate now.

mattyg (changes requested):
Requested some changes inline.

One gotcha I want to make sure is handled (it may be handled already but we should confirm).

Say we have a private conversation with a contact, but we never received a profile for that contact agent and conversation.  Then we delete the contact. What will be displayed as the profile name in the conversation messages, or the profile in the ConversationSummary, or in the title from the ConversationTitleStore?

ui/src/lib/svgIcons.ts
MasterK0927: Fill="none" Was required for this icon

mattyg: Why? Is it because the icon is all stroke and no fill? If so stroke should be "currentColor" so that it can be modified via the css text color of the parent element, and the documentation at the top of the page needs to be updated to reflect this.

As I said I want a consistent solution for all svgs that is clearly documented as a holdover before getting to #405. Otherwise it will get out of hand messy again quickly.

MasterK0927: Yep, the icon is all stroke

MasterK0927: Putting stroke="currentColor" makes it adaptive to the background but it doesn't matches the red accent of add(+) button.

MasterK0927:
> Requested some changes inline.
> 
> One gotcha I want to make sure is handled (it may be handled already but we should confirm).
> 
> Say we have a private conversation with a contact, but we never received a profile for that contact agent and conversation. Then we delete the contact. What will be displayed as the profile name in the conversation messages, or the profile in the ConversationSummary, or in the title from the ConversationTitleStore?

As per my understanding, we need to modify the delete behavior to maintain a "tombstone" record with basic display information even after deletion similar to how social media platforms handle deleted accounts.

Let me explain it in more detail.

Currently, when a contact is deleted, two key things happen:

- Contact is removed from the contacts store.
- `cellId` mapping for that contact is removed.

There is a potential edge case that we are surely missing. While the contact information is removed from the `ContactStore` , the messages in the private conversation would still exist and need to reference the sender's profile information.

If we never received a profile for that contact and then deleted the contact entry, we would lose the connection between the agent's public key and any profile information.

When trying to display profile information in conversation messages, `ConversationSummary`, or `ConversationTitleStore`, we would likely encounter one of these scenarios:

1. **Looking up profile information through the contact store**  
   The components would find nothing since the contact was deleted.

2. **Fetching profile information directly from the DHT**  
   There would be nothing to find since we never received a profile.

To properly handle this case, we could consider one of the following solutions:

- Keep a permanent record of basic agent information separate from the contacts store.  
  This could be a lightweight store that maintains a mapping of agent public keys to their last known display information, even after contact deletion.

- Implement fallback display logic in UI components.  
  This would show the agent's public key (truncated) when no profile information is available.

ui/src/lib/ButtonDelete.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+  import SvgIcon from "./SvgIcon.svelte";
+  export let moreClasses: string;
```
mattyg:
```suggestion
  export let moreClasses: string = "";
```

ui/src/lib/ButtonDelete.svelte
```diff
@@ -0,0 +1,13 @@
...
+</script>
+
+<button
+  type="button"
+  on:click|stopPropagation
+  class="inline-flex h-8 w-8 items-center justify-center rounded-full transition-opacity hover:opacity-75 focus:outline-none"
+  aria-label="Delete"
+>
+  <SvgIcon icon="delete" {moreClasses} />
+</button>
```
mattyg: Does it make sense to create a new component versus using the existing ButtonIconBare?

MasterK0927: I tried ButtonIconBare and other Button Types, but was having issues with them.

ui/src/lib/svgIcons.ts
mattyg: Sounds like you'll have to investigate further to find a solution.

mattyg:
> When trying to display profile information in conversation messages, ConversationSummary, or ConversationTitleStore, we would likely encounter one of these scenarios:

Did you actually confirm this is the case? I think we may be handling it because currently messages are not included the readable for agents who do not have a profile. Not necessarily the ideal solution, but not terrible either.

If necessary I would prefer this option:
> Implement fallback display logic in UI components.
> This would show the agent's public key (truncated) when no profile information is available.

ui/src/lib/ButtonDelete.svelte
mattyg: Can you elaborate? What issues?

MasterK0927: Like improper stylings and issues keeping it in line with the existing "+" icon. I had tested every button types and tried to make them align with our need for this delete icon, but none worked perfectly, so rather than using the `button` tag created a new component that can be reused wherever we want to display the delete icon.

MasterK0927:
> > When trying to display profile information in conversation messages, ConversationSummary, or ConversationTitleStore, we would likely encounter one of these scenarios:
> 
> Did you actually confirm this is the case? I think we may be handling it because currently messages are not included the readable for agents who do not have a profile. Not necessarily the ideal solution, but not terrible either.
> 
> If necessary I would prefer this option:
> 
> > Implement fallback display logic in UI components.
> > This would show the agent's public key (truncated) when no profile information is available.

Yeah I am confirming that we have this issue, we are not actually removing the name of the contact deleted from the conversations.

ui/src/lib/ButtonDelete.svelte
mattyg: I don't want us to create a new Button component for every different icon using it. Instead if the existing generic ButtonIconBare doesn't suit your needs, you can either pass in additional css classes to it using the moreClasses props or create another generic button component with your desired styling and pass your icon into it.

It will be a good idea to extract this wrapper of the generic button component into a ButtonDelete component, but lets save that for when we use it more than once.

ui/src/lib/svgIcons.ts
```diff
@@ -90,4 +90,5 @@ export const svgIcons: { [key: string]: string } = {
...
+  delete: '<svg width="100%" height="100%" viewBox="0 0 25 25" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M5 6.66671H20.8333M19.1667 7.50004V19.1667C19.1667 19.8297 18.9033 20.4656 18.4344 20.9345C17.9656 21.4033 17.3297 21.6667 16.6667 21.6667H9.16667C8.50363 21.6667 7.86774 21.4033 7.3989 20.9345C6.93006 20.4656 6.66667 19.8297 6.66667 19.1667V7.50004M10.8333 6.66671V4.16671C10.8333 3.94569 10.9211 3.73373 11.0774 3.57745C11.2337 3.42117 11.4457 3.33337 11.6667 3.33337H14.1667C14.3877 3.33337 14.5996 3.42117 14.7559 3.57745C14.9122 3.73373 15 3.94569 15 4.16671V6.66671M10.8333 10.8334V16.6667M15 10.8334V16.6667" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></svg>',
```
mattyg, line 94: Looks like you still need to update the documentation at the top of the file to reflect this kind of icon.

ui/src/store/RelayClient.ts
```diff
@@ -320,7 +320,7 @@ export class RelayClient {
       cell_id: this.provisionedRelayCellId,
       zome_name: ZOME_NAME,
       fn_name: "create_contact",
-      payload,
+      payload,
```
mattyg:
```suggestion
      payload,
```

ui/src/lib/svgIcons.ts
MasterK0927: Means I have to add exception at documentation for this kind of icons, right?

ui/src/lib/ContactListItem.svelte
```diff
@@ -54,25 +66,34 @@
...
+      {/if}
+    </p>
+  </div>
+  <div class="flex items-center space-x-2">
+    {#if selected}
+      <ButtonInline
+        on:click={() => goto("/contacts/" + $contact.publicKeyB64)}
+        moreClasses="dark:bg-secondary-700 "
+      >
+        {$t("create.view")}
```
mattyg:
```suggestion
        {$t("common.view")}
```

ui/src/lib/ContactListItem.svelte
```diff
@@ -54,25 +66,34 @@
...
+        : ''}"
     >
-      {$t("common.view")}
-    </ButtonInline>
-  {:else}
-    <span class="text-primary-500 text-lg font-extrabold">+</span>
-  {/if}
+      {$contact.fullName}
+      {#if hasAgentJoinedDht}
+        <span class="text-secondary-400 ml-1 text-xs">{$t("create.unconfirmed")}</span>
```
mattyg:
```suggestion
        <span class="text-secondary-400 ml-1 text-xs">{$t("common.unconfirmed")}</span>
```

ui/src/lib/svgIcons.ts
mattyg: Not necessarily an "exception" but update the documentation to reflect the different required changes for a fill-based icon vs a stroke-based icon

mattyg (changes requested):
Thanks, this is looking good. 

I think the UI could be improved a bit. Currently we have these two icons next to each other, which is a bit cluttered: 
image

They're so close together that you might try to tap one and accidentally delete the contact, which is exacerbated by not having a confirmation dialog asking the user to confirm the delete action

So to remedy this I'd like to request the following UI changes:
- Instead of having the delete button be on the ContactListItem, move it onto the /contact/[id] page. It can be a normal Button there where this is more room available.
- Upon clicking the delete button, open a confirmation dialog that says "Are you sure you want to delete this contact?" with two buttons: Cancel and Delete. To implement this, please create a generic Dialog component with prop for the action button label and a slot for the body content.

ui/src/lib/svgIcons.ts
```diff
@@ -7,7 +7,8 @@
  * Before adding an svg string here you MUST complete the following steps:
  * 1. Ensure the svg string does NOT contain the properties: "fill", "stroke", "width", or "height"
  *     If it does, delete them.
- * 2. Add the following properties to the svg element: fill="currentColor" width="100%" height="100%"
+ * 2. For fill based icons, add the following properties to the svg element: fill="currentColor" width="100%" height="100%"
+ * 3. For stroke based icons, add the following properties to the svg element: stroke="currentColor"
```
mattyg: We also need to set `width` and `height` for stroke-based icons as well.

ui/src/store/ContactStore.ts
```diff
@@ -76,6 +77,21 @@ export function createContactStore(client: RelayClient): ContactStore {
...
 
+  /**
+   * Delete a contact
+   */
+  async function deleteContact(agentPubKeyB64: AgentPubKeyB64) {
+    const contact = contacts.getKeyValue(agentPubKeyB64);
+    await client.deleteContact(contact.originalActionHash);
+    // Updates the local stores
+    contacts.removeKeyValue(agentPubKeyB64);
+    cellIds.update((d) => {
```
mattyg, line 86: Please swap the order of these store updates: first delete the cellId, then remove the contact. That way we can't get into a state where a cellId exists for a contact that is was deleted.

ui/src/routes/contacts/[id]/+page.svelte
```diff
@@ -33,85 +33,142 @@
...
+    } finally {
+      isLoading = false;
+    }
+  }
+
+  async function handleDeleteContact() {
+    if (isDeletingContact) return;
+
+    isDeletingContact = true;
+    showDeleteDialog = false;
```
Comment (mattyg) [inline ui/src/routes/contacts/[id]/+page.svelte]: Either handle setting the `showDeleteDialog` manually from the parent, or use the `bind:open` and set open from the child, but using both is redundant.

ui/src/routes/contacts/[id]/+page.svelte
```diff
@@ -33,85 +33,142 @@
...
-      ? deriveCellConversationStore(conversationStore, encodeCellIdToBase64($contact.cellId))
-      : undefined;
-  let profiles =
-    $contact.cellId !== undefined
-      ? deriveCellProfileStore(profileStore, encodeCellIdToBase64($contact.cellId))
-      : undefined;
+  let conversation: ReturnType<typeof deriveCellConversationStore> | undefined;
+  let profiles: ReturnType<typeof deriveCellProfileStore> | undefined;
+
+  $: if ($contact && $contact.cellId) {
```
Comment (mattyg) [inline ui/src/routes/contacts/[id]/+page.svelte]: - We don't need to use `$: ...` derived for this since this is a page component, so every time it is rendered the `let` variables will be instantiated again. 
- We should never be able to access this page if the contact doesn't exist, since there will be no links to it anywhere, so I don't think that needs to be handled. Upon deleting the contact we should be navigated away from the page.

ui/src/routes/contacts/[id]/+page.svelte
```diff
@@ -33,85 +33,142 @@
...
+    showDeleteDialog = false;
+
+    try {
+      await contact.delete();
+      toast.success($t("common.delete_contact_success"));
+      await goto("/create");
+    } catch (err) {
+      console.error("Error deleting contact:", err);
+      toast.error($t("common.delete_contact_error"));
+      showDeleteDialog = false;
```
Comment (mattyg) [inline ui/src/routes/contacts/[id]/+page.svelte]: We already set `showDeleteDialog = false;` above, and in the child component. Need just one implementation for closing the dialog.

ui/src/routes/contacts/[id]/+page.svelte
```diff
@@ -33,85 +33,142 @@
...
+    } finally {
+      isDeletingContact = false;
```
Comment (mattyg) [inline ui/src/routes/contacts/[id]/+page.svelte]: nitpick but we're not using `finally` blocks anywhere else in the codebase so for syntax consistency I would just put this after the try/catch blocks.

ui/src/lib/ContactListItem.svelte
mattyg, line 1: I don't think any of these formatting changes are necessary anymore now that the additional button was removed from this component

mattyg (changes requested):
requested changes in comments inline

ui/src/lib/Dialog.svelte
```diff
@@ -0,0 +1,43 @@
+<script lang="ts">
+  import Button from "$lib/Button.svelte";
+  import { createEventDispatcher } from "svelte";
+  import { t } from "$translations";
+
+  export let open = false;
+  export let title: string;
+  export let actionButtonLabel = $t("common.confirm");
+
+  const dispatch = createEventDispatcher();
```
mattyg:
```suggestion
  const dispatch = createEventDispatcher<{confirm: null, cancel: null}>();
```

ui/src/translations/locales/en.json
```diff
@@ -26,6 +27,12 @@
   "create_group": "Create group",
   "create_new_contact": "Create New Contact",
   "created": "Created: {{date:date;}}",
+  "delete": "Delete",
+  "delete_contact": "Delete Contact",
+  "delete_contact_error": "Error deleting contact",
+  "delete_contact_success": "Contact deleted successfully",
+  "delete_contact_title": "Delete Contact",
```
mattyg: Why do we need the same string twice?

MasterK0927: Ah! Missed it.

ui/src/store/ContactStore.ts
mattyg: Still not resolved

ui/src/routes/contacts/[id]/+page.svelte
```diff
@@ -45,73 +49,121 @@
...
 
+  let showDeleteDialog = false;
+
   let pollInterval: NodeJS.Timeout;
   $: hasAgentJoinedDht =
     profiles !== undefined &&
-    $profiles?.list.find(([key]) => key === $contact.publicKeyB64) !== undefined;
+    $profiles?.list.find(([key]) => key === $contact?.publicKeyB64) !== undefined;
 
   async function loadProfiles() {
```
Comment (mattyg) [inline ui/src/routes/contacts/[id]/+page.svelte:57]: This function is not necessary to complete before rendering the page, its just polling to see if the agent has joined the DHT since we originally loaded the page. Thus we don't need to track isLoading here and we don't need to wait for isLoading before rendering the page.

ui/src/routes/contacts/[id]/+page.svelte
```diff
@@ -45,73 +49,121 @@
       ? deriveCellProfileStore(profileStore, encodeCellIdToBase64($contact.cellId))
       : undefined;
 
+  let showDeleteDialog = false;
+
   let pollInterval: NodeJS.Timeout;
   $: hasAgentJoinedDht =
     profiles !== undefined &&
-    $profiles?.list.find(([key]) => key === $contact.publicKeyB64) !== undefined;
+    $profiles?.list.find(([key]) => key === $contact?.publicKeyB64) !== undefined;
```
Comment (mattyg) [inline ui/src/routes/contacts/[id]/+page.svelte]: Don't think this change is needed either anymore

mattyg (changes requested):
see inline comments

ui/src/routes/contacts/[id]/+page.svelte
Comment (MasterK0927) [reply ui/src/routes/contacts/[id]/+page.svelte]: The `?.` operator is needed here because when a contact is deleted, there's a brief moment where the contact data becomes undefined. Without `?.`, trying to read properties during that moment would cause an error that freezes the UI update.`?.` prevents this error, ensuring the deletion completes smoothly with instant UI feedback.

MasterK0927:
Dialog component in smaller viewports

mattyg (approved):
I made a few minor changes to the implementation:
- The dialog stays open while the contact is being deleted, and the "loading" indicator is shown in the dialog action button
- Show the delete icon to the dialog action button
- explicitly check if $contact is undefined once, rather than within the `find` callback
- Added a period to the end of the string `delete_contact_dialog_message`
- Added the feature to the changelog

Great work!

PR #497, opened 2025-01-23
Feat/delete message

### LINKED ISSUE #177: Delete message
- I can delete any messages I've sent
- Upon clicking 'delete' button a confirmation notice is displayed with text explaining that peers may still hold a copy of your data
- When a message is deleted, a placeholder is displayed noting "This message was deleted". The placeholder should be in the same position in the conversation as the original message. The placeholder should say the deletion time.

Relies on #133

Note you can never actually force another peer to delete some data you published (even with "purge" or dna migration), but this fulfills a similar UX expectation of signalling to others that "I didn't mean to send this message"

MasterK0927:
Resolves #177 

- Adds a `delete` MessageAction button. 
- Shows a confirmation dialog before deletion with proper warning. 
- ~~Adds a placeholder in place of deleted message, which reads as `"Message deleted on ${timestamp}"`.~~


MasterK0927:
@mattyg I need help showing the deleted message placeholder to participants. I’m able to delete the message, but there are currently two major issues:

1. The deleted message is reflected as deleted only on the sender's chat screen, not on the participants. 
2. When the chat is refreshed, the deleted message completely disappears from the chat screen (I mean the placeholder).

I tried changing how we’re currently loading new messages in the chat to include two distinct functions: `_filterMissingMessages` for getting the new message hashes and `_filterDeletedMessagesForLoading` for getting the deleted message hashes. However, I wasn’t able to fix the above two issues, so I need assistance with them.

Additionally, the UX is mostly finished, but some changes are still required.

MasterK0927:
@mattyg I had figured out the solution and implemented the same, now I am able to tackle the issue reported earlier. Its now ready for review ;)

The current implementation of conversations follows the flow below:
`+page.svelte` → `ConversationMessages.svelte` → `BaseMessage` (`Message.svelte`) → `MessageAction.svelte`

To display the Delete button in `MessageAction.svelte` and show the Confirmation Dialog in `+page.svelte`, we need to pass the click event all the way up to the top of the flow, i.e., `+page.svelte`. We have three approaches to achieve this:

1. Use event dispatchers to send events hierarchically through the intermediate components up to the main page. In this case, `ConversationMessages.svelte` acts solely as an event relay and does not perform the actual action, since we cannot bubble events through components  skipping any component in between in Svelte. (I am using this approach.)
2. Pass a delete handler function from `+page.svelte` down through `ConversationMessages` to the `Message` component. This way, when the delete action is triggered in `MessageActions`, it directly calls the handler without multiple event dispatches.
3. Create a separate store to manage the delete dialog state, which is currently handled in `+page.svelte`. By moving it to the store, we can import the store and trigger the dialog and other actions directly. However, this would couple `MessageActions` to the store, which may not be ideal for reusability.


**Current issues:**

- The state of deleted message is getting updated after 2s only, not immediately for all the participants. 
- Upon refreshing the app (relaunching), the deleted message placeholder is disappearing, which I doubt is due to the reason that we are removing the links of the deleted messages.

mattyg:
> To display the Delete button in MessageAction.svelte and show the Confirmation Dialog in +page.svelte, we need to pass the click event all the way up to the top of the flow, i.e., +page.svelte. We have three approaches to achieve this:

I'd say just bubble up the event, or else just render a confirmation dialog inside each MessageAction.svelte

mattyg:
> The state of deleted message is getting updated after 2s only, not immediately for all the participants.

I don't understand this -- are you saying that when *I* delete my message, that it takes a few seconds to appear deleted to me? Or are you saying it takes 2s for other peers to see that I deleted my message?

mattyg:
> Upon refreshing the app (relaunching), the deleted message placeholder is disappearing, which I doubt is due to the reason that we are removing the links of the deleted messages.

We need to use `get_details` requests to fetch the delete actions from the DHT. See https://docs.rs/hdk/latest/hdk/entry/fn.get_details.html and https://docs.rs/hdk/latest/hdk/link/fn.get_link_details.html

I believe we already have zome functions that provide this data, so we may just need to change / add additional zome calls from the front-end.

ui/src/lib/types.ts
```diff
@@ -85,6 +85,11 @@ export interface MessageExtended {
   timestamp: number;
 }
 
+export interface MessageExtendedWithDeletion extends MessageExtended {
+  isDeleted?: boolean;
+  deletedAt?: string;
```
mattyg: these fields are redundant right? If `deletedAt === undefined` that is equivalent to `isDeleted === false`. So we only need `deletedAt`

ui/src/lib/utils.ts
```diff
@@ -141,3 +142,8 @@ export function isWithinFiveMinutes(d1: Date, d2?: Date): boolean {
 
   return Math.abs(d1.getTime() - d2.getTime()) <= 5 * 60 * 1000;
 }
+
+export function formatHolochainTimestamp(timestamp: number): string {
```
mattyg: :+1:

ui/src/lib/types.ts
```diff
@@ -85,6 +85,11 @@ export interface MessageExtended {
   timestamp: number;
 }
 
+export interface MessageExtendedWithDeletion extends MessageExtended {
```
mattyg: Can we just add these fields to the MessageExtended type? The idea of the "*Extended" types are they contain additional data, beyond what we fetched in the DHT get, specifically for rendering the model in the front-end. The goal is that we have one place where these Extended types are constructed, and they are used throughout the front-end.

ui/src/routes/conversations/[id]/MessageActions.svelte
```diff
@@ -1,22 +1,34 @@
...
   import { page } from "$app/stores";
-  import { encodeHashToBase64 } from "@holochain/client";
+  import { encodeHashToBase64, type ActionHashB64, type AgentPubKeyB64 } from "@holochain/client";
   const fileStore = getContext<{
     getStore: () => FileStore;
   }>("fileStore").getStore();
   let cellFileStore = deriveCellFileStore(fileStore, $page.params.id);
 
-  export let message: MessageExtended;
+  const actionHashB64 = getContext<ActionHashB64>("messageActionHash");
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/MessageActions.svelte]: I don't understand this. We already have the MessageExtended which contains the action hash.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +184,91 @@ export function createConversationMessageStore(
...
+  /**
+   * Delete Message by Content, from the content of the message,
+   * and mark it as deleted in the store.
+   *
+   * @param key1 CellIdB64
+   * @param messageContent string
+   * @returns
+   */
+
+  function _createDeletionPlaceholder(
```
mattyg: The deleted message placeholder content should be specified at the component level when we're rendering the message. If the message contains a `deletedAt` value, then we render the deletion placeholder, otherwise we render the message

In the store we can still store the message, just with the difference is a deleted message has a `deletedAt` timestamp.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +184,91 @@ export function createConversationMessageStore(
...
+    const cellId = decodeCellIdFromBase64(key1);
+
+    const existingMessages = get(messages).data[key1] || {};
+    const existingMessage = existingMessages[actionHashB64];
+
+    if (!existingMessage) throw new Error("Message not found");
+
+    await client.deleteMessage(cellId, decodeHashFromBase64(actionHashB64));
+    const deleteStatus = await client.getDeleteStatus(cellId, decodeHashFromBase64(actionHashB64));
+    const deletionTimestamp = formatHolochainTimestamp(deleteStatus.deletedAt);
```
mattyg: The timestamp formatting should be done at the component level, not in the store.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +184,91 @@ export function createConversationMessageStore(
...
+    };
+  }
+
+  async function deleteMessage(key1: CellIdB64, actionHashB64: ActionHashB64): Promise<void> {
+    const cellId = decodeCellIdFromBase64(key1);
+
+    const existingMessages = get(messages).data[key1] || {};
+    const existingMessage = existingMessages[actionHashB64];
+
+    if (!existingMessage) throw new Error("Message not found");
```
mattyg: Checking if we have the message locally is not necessary. Just do the zome call let it error if the message doesn't exist.

ui/src/store/RelayClient.ts
```diff
@@ -341,4 +341,46 @@ export class RelayClient {
...
+      zome_name: ZOME_NAME,
+      fn_name: "delete_message",
+      payload: messageHash,
+    });
+  }
+
+  public async getDeleteStatus(
+    cellId: CellId,
+    messageHash: ActionHash,
+  ): Promise<{
```
mattyg: Just return the raw result from the zome call here. The caller can extract the data it cares about.

ui/src/store/RelayClient.ts
```diff
@@ -341,4 +341,46 @@ export class RelayClient {
...
+
+  /**
+   * Delete a message
+   *
+   * Delete a message by its hash, and get the status of the deletion.
+   * Message is not actually deleted, but a delete action is created.
+   *
+   */
+
+  public async deleteMessage(cellId: CellId, messageHash: ActionHash): Promise<ActionHash> {
```
mattyg:
```suggestion
  public async deleteMessage(cellId: CellId, payload: ActionHash): Promise<ActionHash> {
```

ui/src/store/RelayClient.ts
```diff
@@ -341,4 +341,46 @@ export class RelayClient {
...
+    return this.client.callZome({
+      cell_id: cellId,
+      zome_name: ZOME_NAME,
+      fn_name: "delete_message",
+      payload: messageHash,
```
mattyg:
```suggestion
      payload,
```

nitpick but more legible

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +184,91 @@ export function createConversationMessageStore(
...
+
+    await client.deleteMessage(cellId, decodeHashFromBase64(actionHashB64));
+    const deleteStatus = await client.getDeleteStatus(cellId, decodeHashFromBase64(actionHashB64));
```
mattyg: Getting the delete status of a message can be moved into the `_makeMessageExtended` function, since we need that information for every message before it gets added to the store.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +184,91 @@ export function createConversationMessageStore(
...
+    const deletionTimestamp = formatHolochainTimestamp(deleteStatus.deletedAt);
+
+    const deletedMessageExtended = _createDeletionPlaceholder(
```
mattyg: We don't want a different data structure for deleted messages, it should be a MessageExtended with `deletedAt` defined

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +184,91 @@ export function createConversationMessageStore(
...
+    messages.update((m) => ({
+      ...m,
+      [key1]: {
+        ...(m[key1] || {}),
+        [actionHashB64]: deletedMessageExtended,
+      },
+    }));
+  }
+
+  async function markMessageAsDeleted(key1: CellIdB64, actionHashB64: ActionHashB64) {
```
mattyg: This is only called from an incoming signal. Can we rename it to mirror `handleMessageSignalReceived`, for example `handleMessageDeletedSignalRecieved`.

Its possible we may not need both functions, since `handleMessageSignalReceived` is already calling `makeMessageExtended` which itself is getting the message status. Especially because we don't need to display the message deleted placeholder when received via a signal unless we've already have the original message in the store already (I don't think it offers any UX benefit). If that's the case it would be ideal to only have one.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -273,21 +366,45 @@ export function createConversationMessageStore(
     maxBucketsToFetch?: number,
   ): Promise<number> {
     // Fetch the list of buckets that contain the target count
-    // This step is split out so that buckets can be fetched in chunks, in parallel
```
mattyg: ?

ui/src/store/ConversationMessageStore.ts
```diff
@@ -273,21 +366,45 @@ export function createConversationMessageStore(
...
+          };
+        }
+      }),
+    );
+    messages.update((m) => ({
+      ...m,
+      [key1]: updatedMessages,
+    }));
 
-    // Fetch and save missing message data to store
```
mattyg: ?

ui/src/store/ConversationMessageStore.ts
```diff
@@ -273,21 +366,45 @@ export function createConversationMessageStore(
...
       bucket,
       targetCount,
       bucketChunkSize,
       maxBucketsToFetch,
     );
-    const actionHashB64s = flatten(bucketsToFetch.map(({ actionHashB64s }) => actionHashB64s));
-
-    // Filter only messages we are not storing already
-    const missingActionHashB64s = await _filterMissingMessages(key1, actionHashB64s);
+    const allActionHashB64s = flatten(bucketsToFetch.map(({ actionHashB64s }) => actionHashB64s));
```
mattyg: None of the changes to this function should be necessary, we're already getting the delete status in _makeMessageExtended.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -418,28 +533,42 @@ export function createConversationMessageStore(
...
+        messageRecord.signed_action.hashed.content.timestamp,
+      );
+    }
 
     return {
       message: messageRecord.message,
       authorAgentPubKeyB64: encodeHashToBase64(messageRecord.signed_action.hashed.content.author),
       timestamp: messageRecord.signed_action.hashed.content.timestamp,
+      isDeleted: false,
     };
```
mattyg: Same notes as above:
- MessageExtended adds a `deletedAt?: number` field only
- Deletion placeholder content is determined at the component level when rendering the message

mattyg (changes requested):
lmk if you need any clarification on the inline comments

ui/src/routes/conversations/[id]/MessageActions.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/MessageActions.svelte]: MessageRecord has the actionhash not the messageExtended

MasterK0927:
@mattyg I have incorporated all your suggested changes, but the previous two issues still persist. Additionally, a new issue has emerged: when deleting a message, the placeholder is only visible to the author and not updated for peers (after removing the placeholder adding from store).

After reviewing the documentation you referenced and investigating the handlers in `RelayClient.ts`, I don’t see any issues with `deleteMessage` or `deleteStatus`. While I suspect `getMessageEntries` might be causing the problem, I’m not entirely sure.

I would greatly appreciate your help in the implementation further.

ui/src/store/RelayClient.ts
```diff
@@ -341,4 +342,35 @@ export class RelayClient {
...
+    payload: ActionHash,
+  ): Promise<SignedActionHashed | undefined> {
+    const deletedAction: SignedActionHashed = await this.client.callZome({
+      cell_id: cellId,
+      zome_name: ZOME_NAME,
+      fn_name: "get_oldest_delete_for_message",
+      payload,
+    });
+
+    return deletedAction || undefined;
```
mattyg, line 374: This is not necessary -- the zome call is returning an `Option<SignedActionHashed>` which gets serialized to `SignedActionHashed | undefined`. The previous line is improperly typed.

ui/src/routes/conversations/[id]/MessageActions.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/MessageActions.svelte]: The line below this is `  export let message: MessageExtended;` -- this is the prop that this component takes  as input.

Thus we have the MessageExtended.

Thus we have the action hash which is included in the MessageExtended.

ui/src/routes/conversations/[id]/Message.svelte
```diff
@@ -21,8 +23,12 @@
   export let cellIdB64: CellIdB64;
   export let isSelected: boolean = false;
   export let showAuthor: boolean = false;
+  export let actionHashB64: ActionHashB64;
+
+  setContext("messageActionHash", actionHashB64);
 
   $: fromMe = message.authorAgentPubKeyB64 === myPubKeyB64;
+  $: isDeleted = message.deletedAt;
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/Message.svelte]:
```suggestion
  $: isDeleted = message.deletedAt !== undefined;
```

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -58,16 +59,18 @@
       <BaseMessage
         {cellIdB64}
         message={messageExtended}
-        isSelected={selected === actionHashB64}
+        isSelected={messageExtended.deletedAt ? false : selected === actionHashB64}
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]: I'd find it clearer to change the logic of the `handleClick` above to not select messages that are deleted.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -58,16 +59,18 @@
...
         showAuthor={prevMessageExtended === undefined ||
           messageExtended.authorAgentPubKeyB64 !== prevMessageExtended.authorAgentPubKeyB64 ||
           !isWithinFiveMinutes(
             new Date(messageExtended.timestamp / 1000),
             new Date(prevMessageExtended.timestamp / 1000),
           )}
+        {actionHashB64}
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]: We're already passing the messageExtended, so we don't need the actionhashB64 which is included in that

ui/src/store/SignalHandler.ts
```diff
@@ -17,24 +17,29 @@ export function createSignalHandler(
...
+        signal[SignalType.App].payload as MessageSignal,
+      );
+      // Mark conversation as unread
+      // Unless user is currently viewing the conversation page.
+      const $page = get(page);
+      if ($page.params.id !== cellIdB64 || $page.route.id !== "/conversations/[id]") {
+        await conversationStore.updateUnread(cellIdB64, true);
+      }
+    } else if (payload.type === "EntryDeleted" && payload.original_app_entry.type === "Message") {
+      const originalActionHashB64 = payload.action.hashed.hash;
```
mattyg:
```suggestion
      const originalActionHash = payload.action.hashed.hash;
```

Is the actually the original action hash? Or is it the latest action hash (i.e. the delete action)? I think this might be your problem.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +186,58 @@ export function createConversationMessageStore(
...
+          deletedAt: deletionStatus?.hashed.content.timestamp,
+        },
+      },
+    }));
+  }
+
+  async function handleMessageDeletedSignalReceived(key1: CellIdB64, actionHashB64: ActionHashB64) {
+    const cellId = decodeCellIdFromBase64(key1);
+
+    const deletionStatus = await client.getDeleteStatus(
```
mattyg: Isn't the delete action timestamp included in the signal we just received? Why do we need to fetch it again?

ui/src/store/ConversationMessageStore.ts
```diff
@@ -419,28 +481,39 @@ export function createConversationMessageStore(
...
-    messageRecord.message.images.forEach((messageFile) =>
-      fileStore.download(
-        encodeCellIdToBase64(cellId),
-        encodeHashToBase64(messageFile.storage_entry_hash),
-      ),
-    );
+    const deleteStatus = await client.getDeleteStatus(cellId, messageRecord.original_action);
 
-    return {
+    const baseMessage: MessageExtended = {
```
mattyg, line 473:
```suggestion
    const messageExtended: MessageExtended = {
```

ui/src/store/RelayClient.ts
```diff
@@ -341,4 +342,35 @@ export class RelayClient {
...
+      fn_name: "delete_message",
+      payload,
+    });
+  }
+
+  public async getDeleteStatus(
+    cellId: CellId,
+    payload: ActionHash,
+  ): Promise<SignedActionHashed | undefined> {
+    const deletedAction: SignedActionHashed = await this.client.callZome({
```
mattyg:
```suggestion
    const deletedAction: SignedActionHashed | null = await this.client.callZome({
```

mattyg (changes requested):
> but the previous two issues still persist. Additionally, a new issue has emerged: when deleting a message, the placeholder is only visible to the author and not updated for peers (after removing the placeholder adding from store).

Please take a week to try to diagnose and resolve the issues. If you're still stuck then I'll take over the PR.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +186,58 @@ export function createConversationMessageStore(
     }));
   }
 
+  /**
+   * Delete Message by Content, from the content of the message,
```
mattyg:
```suggestion
   * Delete Message
```

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +186,58 @@ export function createConversationMessageStore(
...
+      decodeHashFromBase64(actionHashB64),
+    );
+
+    messages.update((m) => ({
+      ...m,
+      [key1]: {
+        ...m[key1],
+        [actionHashB64]: {
+          ...m[key1]?.[actionHashB64],
+          deletedAt: deletionStatus?.hashed.content.timestamp,
```
mattyg, line 213: We may not have the deleted message that we received a signal for in the store. In that case I would just ignore the deleted signal.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -176,6 +186,58 @@ export function createConversationMessageStore(
...
+        },
```
mattyg: We may not have the deleted message that we received a signal for in the store. In that case lets not update the store.

ui/src/routes/conversations/[id]/MessageActions.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/MessageActions.svelte]: https://github.com/holochain-apps/volla-messages/blob/5d643de87899ea2d8661c643d659b958fbe88a2a/ui/src/lib/types.ts#L82 We don't have `ActionHash` in `MessageExtended` 
 
 https://github.com/holochain-apps/volla-messages/blob/5d643de87899ea2d8661c643d659b958fbe88a2a/ui/src/lib/types.ts#L89 We have `ActionHash` in `MessageRecord`

MasterK0927:
@mattyg the previous 2 issues still persist.

Below is the flow I'm currently adhering to for deleteMessage with a placeholder:

image

I have added logs to debug the flow, comparing what is happening versus what is expected. On the sender’s side, logs indicate that the deletion is processed and the signal is emitted (with logs like “[Receiver] (1) EntryDeleted signal received…”). However, on the actual receiver’s side, no deletion signal logs are being recorded. It appears that remote nodes aren’t receiving or processing the deletion signal.

I suspect the issue is in the red boxed part. Our deletion flow relies solely on post_commit emitting an EntryDeleted signal (without an explicit send_remote_signal call). Should we modify the deletion flow to explicitly emit the signal, similar to create_message, or investigate our relay/capability configuration?

mattyg:
@MasterK0927 and I looked at this together. We just need to modify the `delete_message` coordinator function so that it includes a `remote_signal` function call to notify other peers that the messages has been deleted.

MasterK0927:
@mattyg Its now ready for review! Please have a look at it, so we can have a new release with it.

ui/src/routes/conversations/[id]/MessageActions.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/MessageActions.svelte]: ah sorry my bad

mattyg:
I found some higher level issues with the implementation:
- The "message deleted at X" notice only appeared after immediately deleting the message, not after reloading the page. The behavior should be consistent
- We were refetching messages after deleting one, which wasn't necessary because we were receving the MessageDeleted signal
- We were sending & receiving the MessageDeleted signal to ourselves when we delete our own message. This isn't necessary, we can just wait for the delete_message call to succeed and then remove it from the store

I remember that during a meeting recently we discussed and Dr Wurtzer said that it is not necessary to display any "message deleted at X" placeholder. So rather than fix these issues, I removed the logic to fetch delete message status, and to display the placeholders.

Please take a look at it and approve if you're okay with the changes. You can review just my changes here: https://github.com/holochain-apps/volla-messages/pull/497/files/3be91e609155b4296639d1ec49aabd46ad4cb219..51252a156120f8d9300897bfe77f76a5474bf5b2

mattyg (commented):
see comment

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -64,40 +65,31 @@
   let loadingMessagesNew = false;
   let loadingMessagesOld = false;
 
-  $: iAmProgenitor = $conversation.dnaProperties.progenitor === myPubKeyB64;
-
-  /**
-   * Fetch agent profiles every 2s, until at least 2 profiles are received.
-   */
+  let showDeleteDialog = false;
```
Comment (MasterK0927) [inline ui/src/routes/conversations/[id]/+page.svelte:68]: Got it, by keeping each actions separate rather than nesting everything inside `deleteDialog`, we are optimizing reactivity, ensuring that updating only one specific value doesn’t cause unnecessary re-renders for the entire `deleteDialog` object.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -64,40 +65,31 @@
...
+    if (deleteMessageActionHashB64 === undefined) return;
 
-  async function confirmDelete() {
-    deleteDialog.loading = true;
+    isDeletingMessage = true;
     try {
-      await messages.deleteMessage($page.params.id, deleteDialog.messageHash);
+      await messages.deleteMessage($page.params.id, deleteMessageActionHashB64);
       toast.success($t("common.delete_message_success"));
-      // If the message was deleted,
```
Comment (MasterK0927) [inline ui/src/routes/conversations/[id]/+page.svelte]: Oh, Its handled by the `deleteMessage` method itself.

ui/src/store/ConversationMessageStore.ts
```diff
@@ -211,42 +198,25 @@ export function createConversationMessageStore(
...
-        ...m[key1],
-        [actionHashB64]: {
-          ...m[key1]?.[actionHashB64],
-          deletedAt: deletionTimestamp,
-        },
-      },
+      [key1]: Object.fromEntries(
+        Object.entries(m[key1]).filter(([key2]) => key2 !== actionHashB64),
+      ),
     }));
```
MasterK0927: Wouldn't below will be more memory efficient? If I am not wrong, current impl has a time complexity of O(n), as it loops through all entries before reconstructing the object. 

Below impl will reduce it to O(1).
```suggestion
      messages.update((m) => {
      const k = { ...m[key1] };
      delete k[actionHashB64];
      return {
        ...m,
        [key1]: k,
      };
    });
```

MasterK0927 (commented):
@mattyg please have a look at the inline comment

ui/src/store/ConversationMessageStore.ts
mattyg: Yes good catch -- I actually made that change debugging something else but it wasn't necessary. Using `delete` isn't reactive inside svelte components, but its fine inside the svelte/store update fn. I just pushed the change.

ui/src/routes/conversations/[id]/+page.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/+page.svelte:68]: Hmm, I'm actually not sure if that's the case, though you may be right. I more just wanted consistency with the other places where we've used the ConfirmDialog.

MasterK0927:
@mattyg forgot to add changelog for it, updated the changelog in backport, but need to update the changelog in develop too.

mattyg:
> @mattyg forgot to add changelog for it, updated the changelog in backport, but need to update the changelog in develop too.

cool -- please make a PR with only the changelog update and we'll merge it into develop only

PR #502, opened 2025-02-05
Feat/virtual scrolling

### LINKED ISSUE #494: App crashing regularly, until cache is cleared.
Reported by a user in telegram users group:

> 0.7.5 crashed at least 10 times within an hour
> But clearing cache helps

Image

MasterK0927:
Partially Resolves #494

Implements Virtual Scrolling, which limits the number of messages rendered in the DOM at a time. This optimization addresses part of the issue described [here](https://github.com/holochain-apps/volla-messages/issues/494#issuecomment-2627447652), helping to prevent crashes in the Volla Messages app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,5 @@
     "tslib": "^2.6.2",
     "typescript": "^5.4.4"
   },
-  "dependencies": {
-    "svelte-virtual-scroll-list": "^1.3.0"
-  }
+  "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -46,5 +46,7 @@
     "tslib": "^2.6.2",
     "typescript": "^5.4.4"
   },
-  "dependencies": {}
+  "dependencies": {
+    "svelte-virtual-scroll-list": "^1.3.0"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -45,6 +45,5 @@
     "svelte-preprocess": "^5.1.3",
     "tslib": "^2.6.2",
     "typescript": "^5.4.4"
-  },
-  "dependencies": {}
+  }
 }
--- a/ui/package.json
+++ b/ui/package.json
@@ -42,6 +42,7 @@
     "svelte-french-toast": "^1.2.0",
     "svelte-gestures": "^5.0.6",
     "svelte-qrcode-image": "^2.0.0-alpha.1",
+    "svelte-virtual-scroll-list": "^1.3.0",
     "sveltekit-i18n": "^2.4.2",
     "uuid": "^10.0.0"
   },
@@ -82,7 +83,8 @@
     "typescript": "^5.7.2",
     "typescript-eslint": "^8.16.0",
     "vite": "^5.3.4",
-    "vite-plugin-tailwind-purgecss": "^0.3.3"
+    "vite-plugin-tailwind-purgecss": "^0.3.3",
+    "svelte-virtual-scroll-list": "^1.3.0"
   },
   "type": "module"
 }
--- a/ui/package.json
+++ b/ui/package.json
@@ -23,6 +23,7 @@
     "@lightningrodlabs/we-applet": "^0.17.0-dev.2",
     "@msgpack/msgpack": "^3.0.0-beta2",
     "@svelte-put/clickoutside": "^3.0.2",
+    "@tanstack/svelte-virtual": "^3.13.0",
     "@tauri-apps/api": "^2.0.3",
     "@tauri-apps/plugin-barcode-scanner": "^2.0.0",
     "@tauri-apps/plugin-clipboard-manager": "^2.0.0",
@@ -42,7 +43,6 @@
     "svelte-french-toast": "^1.2.0",
     "svelte-gestures": "^5.0.6",
     "svelte-qrcode-image": "^2.0.0-alpha.1",
-    "svelte-virtual-scroll-list": "^1.3.0",
     "sveltekit-i18n": "^2.4.2",
     "uuid": "^10.0.0"
   },
@@ -83,8 +83,7 @@
     "typescript": "^5.7.2",
     "typescript-eslint": "^8.16.0",
     "vite": "^5.3.4",
-    "vite-plugin-tailwind-purgecss": "^0.3.3",
-    "svelte-virtual-scroll-list": "^1.3.0"
+    "vite-plugin-tailwind-purgecss": "^0.3.3"
   },
   "type": "module"
 }
--- a/ui/src/lib/types.ts
+++ b/ui/src/lib/types.ts
@@ -254,22 +254,7 @@ export interface FileExtended {
   status: FileStatus;
 }
 
-/**
- * Represents sizing data for dynamically measured elements in virtualized lists.
- *
- * Used for:
- * - Caching heights to reduce re-measurements
- * - Tracking whether sizes are measured or estimated
- * - Adjusting scroll positions smoothly as content loads/changes
- *
- * Works with dynamic content where:
- * - Initial sizes are estimated
- * - Actual measurements happen asynchronously
- * - Sizes can change with content updates and resizing
- *
- * @property height - Current height in pixels
- * @property measured - True if height is from actual measurement, false if estimated
- */
+// Represents sizing data for dynamically measured elements in virtualized lists.
 export interface SizeInfo {
   height: number;
   measured: boolean;
--- a/ui/src/lib/types.ts
+++ b/ui/src/lib/types.ts
@@ -254,7 +254,8 @@ export interface FileExtended {
   status: FileStatus;
 }
 
-// Represents sizing data for dynamically measured elements in virtualized lists.
+// Size of DOM element
+// Used when dynamically measuring DOM element included virtual scrolling list
 export interface SizeInfo {
   height: number;
   measured: boolean;
--- a/ui/src/lib/types.ts
+++ b/ui/src/lib/types.ts
@@ -254,6 +254,22 @@ export interface FileExtended {
   status: FileStatus;
 }
 
+/**
+ * Represents sizing data for dynamically measured elements in virtualized lists.
+ *
+ * Used for:
+ * - Caching heights to reduce re-measurements
+ * - Tracking whether sizes are measured or estimated
+ * - Adjusting scroll positions smoothly as content loads/changes
+ *
+ * Works with dynamic content where:
+ * - Initial sizes are estimated
+ * - Actual measurements happen asynchronously
+ * - Sizes can change with content updates and resizing
+ *
+ * @property height - Current height in pixels
+ * @property measured - True if height is from actual measurement, false if estimated
+ */
 export interface SizeInfo {
   height: number;
   measured: boolean;
--- a/ui/src/lib/types.ts
+++ b/ui/src/lib/types.ts
@@ -253,3 +253,8 @@ export interface FileExtended {
   file?: File;
   status: FileStatus;
 }
+
+export interface SizeInfo {
+  height: number;
+  measured: boolean;
+}
--- a/ui/src/routes/conversations/[id]/+page.svelte
+++ b/ui/src/routes/conversations/[id]/+page.svelte
@@ -73,10 +73,8 @@
 
   // Reactive update to scroll to the bottom every time the messages update,
   // but only if the user is near the bottom already
-  $: if ($messages.count > 0) {
-    if (scrollAtBottom) {
-      scrollToBottom(100);
-    }
+  $: if ($messages.count > 0 && scrollAtBottom) {
+    scrollToBottom(100);
   }
 
   /**
--- a/ui/src/routes/conversations/[id]/+page.svelte
+++ b/ui/src/routes/conversations/[id]/+page.svelte
@@ -29,6 +29,7 @@
   } from "$store/MergedProfileContactInviteJoinedStore";
   import { POLLING_INTERVAL_FAST, POLLING_INTERVAL_SLOW } from "$config";
   import SvgIcon from "$lib/SvgIcon.svelte";
+  import { SCROLL_BOTTOM_THRESHOLD, SCROLL_TOP_THRESHOLD } from "$config";
 
   const conversationStore = getContext<{ getStore: () => ConversationStore }>(
     "conversationStore",
@@ -67,9 +68,6 @@
   let sending = false;
   let loadingMessagesNew = false;
   let loadingMessagesOld = false;
-
-  const SCROLL_BOTTOM_THRESHOLD = 100; // How close to the bottom must the user be to consider it "at the bottom"
-  const SCROLL_TOP_THRESHOLD = 300; // How close to the top must the user be to consider it "at the top"
 
   $: iAmProgenitor = $conversation.dnaProperties.progenitor === myPubKeyB64;
 
--- a/ui/src/routes/conversations/[id]/+page.svelte
+++ b/ui/src/routes/conversations/[id]/+page.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import { debounce } from "lodash-es";
   import { type AgentPubKeyB64 } from "@holochain/client";
-  import { getContext, onDestroy, onMount } from "svelte";
+  import { getContext, onDestroy, onMount, tick } from "svelte";
   import { page } from "$app/stores";
   import { goto } from "$app/navigation";
   import Header from "$lib/Header.svelte";
@@ -29,6 +29,7 @@
   } from "$store/MergedProfileContactInviteJoinedStore";
   import { POLLING_INTERVAL_FAST, POLLING_INTERVAL_SLOW } from "$config";
   import SvgIcon from "$lib/SvgIcon.svelte";
+  import VirtualScroll from "svelte-virtual-scroll-list";
 
   const conversationStore = getContext<{ getStore: () => ConversationStore }>(
     "conversationStore",
@@ -67,6 +68,7 @@
   let sending = false;
   let loadingMessagesNew = false;
   let loadingMessagesOld = false;
+  let virtualList: VirtualScroll;
 
   const SCROLL_BOTTOM_THRESHOLD = 100; // How close to the bottom must the user be to consider it "at the bottom"
   const SCROLL_TOP_THRESHOLD = 300; // How close to the top must the user be to consider it "at the top"
@@ -150,6 +152,13 @@
     loadingMessagesOld = true;
     try {
       await messages.loadMessagesInPreviousBucketTargetCount();
+      await tick();
+      if (virtualList) {
+        const prevHeight = virtualList.getScrollSize();
+        await messages.loadMessagesInPreviousBucketTargetCount();
+        await tick();
+        virtualList.scrollToOffset(virtualList.getScrollSize() - prevHeight);
+      }
     } catch (e) {
       console.error(e);
     }
@@ -190,8 +199,7 @@
 
   function scrollToBottom(delay: number = 0) {
     setTimeout(() => {
-      if (!conversationContainerRef) return;
-      conversationContainerRef.scrollTop = conversationContainerRef.scrollHeight;
+      if (virtualList) virtualList.scrollToBottom();
       scrollAtBottom = true;
     }, delay);
   }
@@ -230,8 +238,6 @@
     clearTimeout(agentTimeout);
     clearTimeout(configTimeout);
     clearTimeout(messageTimeout);
-
-    conversationMessageStore.cleanupOlderMessages($page.params.id, 1);
 
     conversationContainerRef.removeEventListener("scroll", handleScroll);
 
@@ -295,7 +301,12 @@
           <SvgIcon icon="spinner" moreClasses="!h-4 mt-4" />
         </div>
       {/if}
-      <ConversationMessages cellIdB64={$page.params.id} messages={$messages.list} />
+      <ConversationMessages
+        bind:virtualList
+        cellIdB64={$page.params.id}
+        messages={$messages.list}
+        on:load-previous={loadMessagesInPreviousBucket}
+      />
       {#if loadingMessagesNew}
         <div class="flex items-center justify-center">
           <SvgIcon icon="spinner" moreClasses="!h-4 mb-4" />
--- a/ui/src/routes/conversations/[id]/+page.svelte
+++ b/ui/src/routes/conversations/[id]/+page.svelte
@@ -231,6 +231,8 @@
     clearTimeout(configTimeout);
     clearTimeout(messageTimeout);
 
+    conversationMessageStore.cleanupOlderMessages($page.params.id, 1);
+
     conversationContainerRef.removeEventListener("scroll", handleScroll);
 
     window.removeEventListener("resize", debouncedHandleResize);
--- a/ui/src/routes/conversations/[id]/+page.svelte
+++ b/ui/src/routes/conversations/[id]/+page.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import { debounce } from "lodash-es";
   import { type AgentPubKeyB64 } from "@holochain/client";
-  import { getContext, onDestroy, onMount, tick } from "svelte";
+  import { getContext, onDestroy, onMount } from "svelte";
   import { page } from "$app/stores";
   import { goto } from "$app/navigation";
   import Header from "$lib/Header.svelte";
@@ -29,7 +29,6 @@
   } from "$store/MergedProfileContactInviteJoinedStore";
   import { POLLING_INTERVAL_FAST, POLLING_INTERVAL_SLOW } from "$config";
   import SvgIcon from "$lib/SvgIcon.svelte";
-  import VirtualScroll from "svelte-virtual-scroll-list";
 
   const conversationStore = getContext<{ getStore: () => ConversationStore }>(
     "conversationStore",
@@ -68,7 +67,6 @@
   let sending = false;
   let loadingMessagesNew = false;
   let loadingMessagesOld = false;
-  let virtualList: VirtualScroll;
 
   const SCROLL_BOTTOM_THRESHOLD = 100; // How close to the bottom must the user be to consider it "at the bottom"
   const SCROLL_TOP_THRESHOLD = 300; // How close to the top must the user be to consider it "at the top"
@@ -152,12 +150,6 @@
     loadingMessagesOld = true;
     try {
       await messages.loadMessagesInPreviousBucketTargetCount();
-      await tick();
-      if (!virtualList) return;
-      const prevHeight = virtualList.getScrollSize();
-      await messages.loadMessagesInPreviousBucketTargetCount();
-      await tick();
-      virtualList.scrollToOffset(virtualList.getScrollSize() - prevHeight);
     } catch (e) {
       console.error(e);
     }
@@ -198,8 +190,8 @@
 
   function scrollToBottom(delay: number = 0) {
     setTimeout(() => {
-      if (!virtualList) return;
-      virtualList.scrollToBottom();
+      if (!conversationContainerRef) return;
+      conversationContainerRef.scrollTop = conversationContainerRef.scrollHeight;
       scrollAtBottom = true;
     }, delay);
   }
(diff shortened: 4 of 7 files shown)


MasterK0927:
**Note:**

Although I have stress-tested it in various ways, we cannot be certain that it fully resolves issue #494. Further testing by beta testers is necessary for verification.

package.json
```diff
@@ -46,5 +46,7 @@
     "tslib": "^2.6.2",
     "typescript": "^5.4.4"
   },
-  "dependencies": {}
+  "dependencies": {
+    "svelte-virtual-scroll-list": "^1.3.0"
```
mattyg: This belongs in the `ui` workspace's package.json, as it is a dependency for the ui.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -150,6 +152,13 @@
     loadingMessagesOld = true;
     try {
       await messages.loadMessagesInPreviousBucketTargetCount();
+      await tick();
+      if (virtualList) {
+        const prevHeight = virtualList.getScrollSize();
+        await messages.loadMessagesInPreviousBucketTargetCount();
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]: Why are we loading messages twice in this single function call?

package.json
mattyg: Also a note that this package has not been updated in 2 years. It seems abandoned. Are there other options that are actively maintained?

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -190,8 +199,7 @@
 
   function scrollToBottom(delay: number = 0) {
     setTimeout(() => {
-      if (!conversationContainerRef) return;
-      conversationContainerRef.scrollTop = conversationContainerRef.scrollHeight;
+      if (virtualList) virtualList.scrollToBottom();
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]:
```suggestion
      if (!virtualList) return;
      virtualList.scrollToBottom();
```

I prefer this pattern because it makes clear that we *always* expect virtuallist to be defined by the time this function is run.

ui/src/routes/conversations/[id]/+page.svelte
```diff
@@ -150,6 +152,13 @@
     loadingMessagesOld = true;
     try {
       await messages.loadMessagesInPreviousBucketTargetCount();
+      await tick();
+      if (virtualList) {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/+page.svelte]:
```suggestion
      if (!virtualList) return
```
I prefer this pattern because it makes clear that we always expect virtuallist to be defined by the time this function is run.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -3,12 +3,59 @@
...
+
+  const dispatch = createEventDispatcher();
 
   export let messages: [ActionHashB64, MessageExtended][];
   export let cellIdB64: CellIdB64;
+  export let virtualList: VirtualScroll;
 
   let selected: ActionHashB64 | undefined;
 
+  function processMessages(messages: [ActionHashB64, MessageExtended][]) {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]: I don't like this approach for a few reasons:
1. A list should ideally contain all elements of a single type. This change has it contain multiple types that we then duck-type to determine what to do with each element. It makes the code far less clear and legible.
2. There is no need to do the processing of whether the current date should be displayed up front. It can be determined at the time the element is rendered as it was before.
3. Displaying the current date will be determined based on the contents of the currently visible elements, so it actually doesn't make sense to determine that outside of the virtual list. (i.e. if message A is the first message we're displaying in the virtual list, then we *do* need to display the date before it, even if there are several messages from the same author before it that we are *not* displaying in the virtual list)

mattyg (changes requested):
see inline comments

package.json
MasterK0927: A notable actively maintained project that implements virtual scrolling in Svelte is [svelte-tiny-virtual-list](https://github.com/jonasgeiler/svelte-tiny-virtual-list)

ui/src/routes/conversations/[id]/+page.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/+page.svelte]: The two calls ensure that (a) the needed messages are loaded and (b) change in content height to adjust the scroll position is correctly calculated, keeping the user’s view consistent. It helps in preserving the scroll position while loading older messages.

Comment (mattyg) [reply ui/src/routes/conversations/[id]/+page.svelte]: Please find a way to achieve a consistent content height without making the same call twice. These are expensive calls fetching data, there's no reason to do them twice.

ui/package.json
```diff
@@ -82,7 +83,8 @@
     "typescript": "^5.7.2",
     "typescript-eslint": "^8.16.0",
     "vite": "^5.3.4",
-    "vite-plugin-tailwind-purgecss": "^0.3.3"
+    "vite-plugin-tailwind-purgecss": "^0.3.3",
+    "svelte-virtual-scroll-list": "^1.3.0"
```
mattyg: This dependency is included twice

ui/src/routes/conversations/[id]/+page.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/+page.svelte]: I have the other approach with me, will be commiting it then

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -39,36 +73,62 @@
 </script>
 
 <div class="flex w-full flex-1 flex-col-reverse p-4">
-  <ul>
-    {#each messages as [actionHashB64, messageExtended], i (actionHashB64)}
-      {@const prevMessageExtended = i === 0 ? undefined : messages[i - 1][1]}
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte:44]: This code that already existed was clear and legible. Just move it into the virtual list component slot.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -39,36 +73,62 @@
...
+    {getSize}
+    keeps={30}
+    topThreshold={300}
+    bottomThreshold={100}
+    on:top={() => dispatch("top")}
+    on:bottom={() => dispatch("bottom")}
+    let:data
+    let:index
+  >
+    {#if (() => {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]: I didn't mean just move the function inside the component (this is a very weird illegible pattern), I meant use the coe that already existed where we're using svelte conditionals to render the header, but jut move it inside the virtualscroll slot.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -39,36 +73,62 @@
...
+        </div>
+      </li>
+    {/if}
+    <BaseMessage
+      {cellIdB64}
+      message={data[1]}
+      isSelected={selected === data[0]}
+      showAuthor={index === 0
+        ? true
+        : (() => {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]: Please don't define a function and immediately executing it. There's no need for this -- the previous implementation was clear and just needed to be moved into the virtuallist slot

package.json
mattyg: That looks better. My recommendation would be to look into the well maintained and used tanstack package: https://github.com/tanstack/virtual

MasterK0927: I am migrating the code to use this library then, it will then take me a day or two for impl and testing it thoroughly so that I can cover all the edge cases.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -1,13 +1,110 @@
...
+      return (showDate ? 40 : 0) + 100;
+    },
+    overscan: 10,
+
+    /**
+     * Measures actual element height and updates size cache
+     * @param element The rendered DOM element
+     * @returns Measured height of the element in pixels
+     */
+    measureElement: (element: HTMLElement) => {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte:75]: This is never actually executed as far as I can tell.

ui/src/lib/types.ts
```diff
@@ -253,3 +253,8 @@ export interface FileExtended {
   file?: File;
   status: FileStatus;
 }
+
+export interface SizeInfo {
```
mattyg, line 259: Please give a comment to explain this and how its used, since its so different from other types.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -1,13 +1,110 @@
...
   export let messages: [ActionHashB64, MessageExtended][];
   export let cellIdB64: CellIdB64;
 
   let selected: ActionHashB64 | undefined;
+  let virtualListEl: HTMLDivElement;
+  let resizeObserver: ResizeObserver;
+  let scrollTimeout: NodeJS.Timeout;
+  let autoScroll = true;
+
+  const SCROLL_BOTTOM_THRESHOLD = 100;
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]: This config variable is now defined both here and in the parent. I suggest we pull them both out into the `config.ts` file so it is only defined once.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -39,36 +136,52 @@
...
           )}
-        on:press={() => handlePress(actionHashB64)}
-        on:click={(e) => handleClick(e, actionHashB64)}
-        on:clickoutside={handleClickOutside}
-      />
-    {/each}
-  </ul>
+
+        <div
+          data-index={virtualItem.index}
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]: Please svelte conventions like `bind:this={myelements[index]}` rather than html `data-` attributes

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -39,36 +136,52 @@
...
-          messageExtended.authorAgentPubKeyB64 !== prevMessageExtended.authorAgentPubKeyB64 ||
-          !isWithinFiveMinutes(
+  <div class="scroll-container" bind:this={virtualListEl} on:scroll={handleScroll}>
+    <div style="position: relative; height: {$virtualizer.getTotalSize()}px; width: 100%;">
+      {#each $virtualizer.getVirtualItems() as virtualItem (messages[virtualItem.index][0])}
+        {@const [actionHashB64, messageExtended] = messages[virtualItem.index]}
+        {@const prevMessage =
+          virtualItem.index > 0 ? messages[virtualItem.index - 1][1] : undefined}
+        {@const showDate =
+          !prevMessage ||
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]:
```suggestion
          prevMessageExtended === undefined ||
```

Prefer explicit undefined comparisons over relying on js "truthiness" which has weird quirks.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -39,36 +136,52 @@
...
-        message={messageExtended}
-        isSelected={selected === actionHashB64}
-        showAuthor={prevMessageExtended === undefined ||
-          messageExtended.authorAgentPubKeyB64 !== prevMessageExtended.authorAgentPubKeyB64 ||
-          !isWithinFiveMinutes(
+  <div class="scroll-container" bind:this={virtualListEl} on:scroll={handleScroll}>
+    <div style="position: relative; height: {$virtualizer.getTotalSize()}px; width: 100%;">
+      {#each $virtualizer.getVirtualItems() as virtualItem (messages[virtualItem.index][0])}
+        {@const [actionHashB64, messageExtended] = messages[virtualItem.index]}
+        {@const prevMessage =
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]:
```suggestion
        {@const prevMessageExtended =
```

The rest of the codebase has been very consistent in variable naming to make clear when we're using *Extended types. Let's keep it that way.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -39,36 +136,52 @@
...
+          virtualItem.index > 0 ? messages[virtualItem.index - 1][1] : undefined}
+        {@const showDate =
+          !prevMessage ||
+          !isSameDay(
             new Date(messageExtended.timestamp / 1000),
-            new Date(prevMessageExtended.timestamp / 1000),
+            new Date(prevMessage?.timestamp / 1000),
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]:
```suggestion
            new Date(prevMessageExtended.timestamp / 1000),
```

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -39,36 +136,52 @@
...
+                })}
+              </div>
+            </li>
+          {/if}
+
+          <BaseMessage
+            {cellIdB64}
+            message={messageExtended}
+            isSelected={selected === actionHashB64}
+            showAuthor={prevMessage === undefined ||
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]:
```suggestion
            showAuthor={prevMessageExtended === undefined ||
```

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -39,36 +136,52 @@
...
+              messageExtended.authorAgentPubKeyB64 !== prevMessage.authorAgentPubKeyB64 ||
+              !isWithinFiveMinutes(
+                new Date(messageExtended.timestamp / 1000),
+                new Date(prevMessage.timestamp / 1000),
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte]:
```suggestion
                new Date(prevMessageExtended.timestamp / 1000),
```

mattyg (changes requested):
Some inline comments. I also noticed that scrolling to the end of the list no longer triggers calling `loadMessagesInCurrentBucket`.


There is definitely some funky implementation in here around handling element heights in a virtual scroll, but I recognize that will be necessary to some extent with virtual scrolling.

I can see how buttery smooth and fast virtual scrolling is, and now we're constraining the message elements that will get rendered to the dom, which will be very helpful specifically when:
- The conversation has a ton of messages
- The user scrolls back in time very far, fetching many pages of messages into the store and rendering them

I'm not convinced this will actually help #494 (I noted it in the issue as a *potential* cause). Can you please reproduce the original issue and then see if this code changes resolves it?

If it does resolve the original issue, then great! If not, I might be inclined to leave this PR open for later to merge when needed, rather than introduce this additional complexity that doesn't actually solve a problem we're facing right now. We may end up needing to do different refactoring that would make this irrelevant (i.e. infinite scroll where messages are loaded into the store by page, and removed from the store by page, when they are scrolled into and out of view).

MasterK0927:
> Some inline comments. I also noticed that scrolling to the end of the list no longer triggers calling `loadMessagesInCurrentBucket`.
> 
> 
> There is definitely some funky implementation in here around handling element heights in a virtual scroll, but I recognize that will be necessary to some extent with virtual scrolling.
> 
> I can see how buttery smooth and fast virtual scrolling is, and now we're constraining the message elements that will get rendered to the dom, which will be very helpful specifically when:
> - The conversation has a ton of messages
> - The user scrolls back in time very far, fetching many pages of messages into the store and rendering them
> 
> I'm not convinced this will actually help #494 (I noted it in the issue as a *potential* cause). Can you please reproduce the original issue and then see if this code changes resolves it?
> 
> If it does resolve the original issue, then great! If not, I might be inclined to leave this PR open for later to merge when needed, rather than introduce this additional complexity that doesn't actually solve a problem we're facing right now. We may end up needing to do different refactoring that would make this irrelevant (i.e. infinite scroll where messages are loaded into the store by page, and removed from the store by page, when they are scrolled into and out of view).


Just reproduced #494 again on an android device running v0.7.5 of Volla Messages. Here is the relevant crash log

Steps for reproducing (worked well in my case) 
- Open Volla Messages App
- Join Volla User Community and wait for 5 minutes for sync
- Move back and forth between chats
- Again go to Volla User Community, it will eventually crash

```
Timestamp: 2025-02-11 09:51:12.736403715+0530 Process uptime: 105s

Zygote Pid: 801

Cmdline: com.volla.messages pid: 5658, tid: 5742, name: ThreadPoolForeg com.volla.messages <<<<

uid: 10630

signal 6 (SIGABRT), code-1 (SI_QUEUEI, fault addr

x2 0000000000000006 x3 0000007b191c2070 x4 00000000ebad808a x5 00000000bad808a x6

00000000ebad808a x7 00000000ebad808b *8 00000000000000F0 x9 0000007c56a62d30

x10 0000000000000001 x11 0000007c56ad79c8

x12 0000000071131000 x13 0000000000b2d888

x14 000000006474e550 x15 0000000000000018

x16 0000007c56b42a80 x17 0000007c56b1b650

x18 0000007617168000 x19 000000000000161a

x20 000000000000166e x21 00000000ffffffff x22

0000000000000000 x23 0000007b33cabc70

x24 0000007b191c23a8 x25.000000009e39e538

x26 0000000014560560 x27 0000000014648690 x28 00000076191c2db4 x29 0000007b191c20f0

Ir 0000007c56ac8948 sp 0000007b191c2050 pc 0000007c56ac8974 pst 0000000000001000 backtrace:

#00 pc 000000000008 d974 /apex/

com.android.runtime/lib64/bionic/libc.so (abort+168) (Buildld: Ba227758540106103d671ea1f80led52)

#01 pc 0000000004032120/data/app/~~e19ssNJQEhVa4QkLPK48-A==/

com.volla.messages-ERFhRuHYxOxTOugxrO1E9Q==/hase.apk Istd::sys: pal.unix: abort_internal::h97e5a2db4196391c+

81

#02 pc 0000000004026714/data/

app/~~e19ssNJQEhVa40kLPK48-A==/

com.volla.messages-ERFhRuHYxOxTougxrO1E9Q==/base.apk

Irust_panic+92)

#03 pc 000000000402b3bc /data/

app/--e19ssNJQEhVa4QkLPK48-A==/

com.volla.messages-ERfhRuHYxOxTOugxrO1E9Q==/hase.apk

Istd::panicking: rust_panic_with_hook::he61021758eba5 700+6081

#04 pc 0000000004026124/data/

app/~~e19ssNJQEhVa40kL.PK48-A==/com.volla.messages-ERFhRuHYxOxT0ugxr01E9Q==/

base.apk Istd::panicking: begin_panic_handler::$u7b$ $u7b$closure$u7d$$u7d$: hd87b429290836638+144)

#05 pc 0000000004029004 /data/

app/~~e19ssNJQEhVa4QkLPK48-A==/

com.volla.messages-ERFhRuHYxOxTOugxrO1E9Q==/base.apk Istd::sys_common::backtrace::__rust_end_short_backtr

ace::h47e2ae5399d92bab+81

#06 pc 000000000402ae98/data/

app/~~e19ssNJQEhVa4QkLPK48-A==/

com.volla.messages-ERfhRuHYxOxTougxrO1E9Q==/base.apk

frust_begin_unwind+481

#07 pc 0000000004044d50 /data/

app/~~e19ssNJQEhVa4QkLPK48-A==/com.volla.messages-ERFhRuHYxOxTougxrO1E9Q==/base.apk (core: panicking: panic_fmt::hf86b5f3919bf49e9+40)

#08 pc 0000000004042460 /data/app/~~e19ssNJOEhVa4QkLPK48-A==/com.volla.messages-ERFhRuHYxOxTougxr01E9Q==/base.apk

(core::cell::panic_already_mutably_borrowed:h3f890f30

38eb88ad+641 #09 pc 0000000003704de0/data/

app/~~e19ssNJQEhVa40kLPK48-A==/

com.volla.messages-ERFhRuHYxOx TOugxrO1E9Q==/base.apk (wry: android::binding: handleRequest: h13d6d4499280dfc

f+24841 #10 pc 000000000031eb8c/data/misc/

apexdata/com.android.art/dalvik-cache/arm64/boot.oat lart_jni_trampoline+1401

#11 pc 0000000002039060/memfd jit-cache (deleted) (com.volla.messages.RustWebViewClient.shouldinte

rceptRequest+528) #12 pc 0000000002038798/memfd:jit-cache Ideleted! (WV.N6.a+8401

#13 pc 0000000000780120/apex/com.android.art/

lib64/libart.so Interp_helper+4016) (Buildid:

c35c9ebf7bb06435e4b31977d87bd5d5)

#14 pc 000000000018c766 /data/

app/--Izbycni 16Vq00FSpYWHsrw==/

com.google.android.webview-pt9hvYnn7ioQsCfjvitYpQ==/base.apk lorg.chromium.android_webview.AwContents Backgr

ound ThreadClient.shouldintercept RequestFromNative+261

#15 pc 000000000036d574 /apex/com.android.art/lib64/libart.so lart_quick_invoke_stub+612) (Buildld:

c35c9ebf7bb06435e4b31977d87bd5d5) #16 pc 000000000036be04 /apex/

com.android.art/lib64/libart.so lart::Value

art: Invoke VirtualOrInterfaceWith VarArgs art::ArtMeth od*>lart::ScopedObjectAccessAlreadyRunnable const&,

_jobject, art::ArtMethod, std::__va_list)+8121) (Buildld: c35c9ebf7bb06435e4b31977d87bd5d5)

#17 pc 000000000074f698 /

apex/com.android.art/lib64/libart.so

lart::INI<false>:: CallObjectMethodVI_JNIEnv*,

_jobject*, _jmethodID*, std::__va_list)+196) (Buildld.

c35c9ebf7bb06435e4b31977d87bd5d5) #18 pc 0000000003200094 /data/app/

~~JQeVQK2J4wnreygioRHgtA==/com.google.android.trichr omelibrary_683416333-Sq2Qz61RSe5PBlq0F06QaQ==/base.apk!libmonochrome_64.so (Buildid:

ac433fc399c6918e19ed2987d803c88499557d3)

#19 pc 00000000048ee72c/data/app/--J0eUQK2j4wnreygioRHatA==/com.google.android.trichr

omelibrary_683416333-Sq2Qz61RSe5PBlq0F060aQ==/base.apk!libmonochrome 64.so (Buildld:

ac433fc399c6918e19ed2987dB 03 c 8 8 4 9 9 557d3f

#20 pc 00000000048ef954 /data/app/--JQeUQK2j4wnreygioRHgtA==/com.google.android.trichro melibrary_683416333-Sq2Qz61RSe5PBly0F06QoQ==/

base.apk!libmonochrome 64.so (Buildld:

ac433fc399c6918e19ed2987dB03c88499557d3fl #21 pc 00000000048efa08 /data/app/

~~JOeUQK2j4wnreygioRHgtA==/com.google.android.trichr omelibrary_683416333-Sq2Qz6IRSe5PBlq0F060aQ/base.apk libmonochrome_64.so (Buildld:

ac433fc399c6918e19ed2987d803c88499557d3f)

#22 pc 00000000042813f8/data/app/~~JOeUQK2j4wnreygioRHgtA==/com.google.android.trichr

omelibrary_683416333-Sq2Qz61RSe5PBlq0F060aQ==/

base.apk!libmonochrome_64.so (Buildld:

ac433fc399c6918e19ed2987d803c88499557d3f) #23 pc 00000000042813ac/data/app/

~~JOeUCK2J4wnreygioRHgtA==/com.google.android.trichro melibrary_683416333-Sq2Qz61RSe5PBlq0F06QaQ==/

base.apk!libmonochrome_64.so (Buildid: ac433fc399c6918e19ed2987d803c88499557d30

#24 pc 0000000003230a74/data/app/

--MeVQK2j4wnreygioRHgtA==/com.google.android.trichro melibrary 683416333-Sq20z61RSe5PBlq0F060aQ==/base.apk libmonochrome_64.so (Buildld:

ac433fc399c6918e19ed2987d803c88499557d3f) #25 pc 000000000322fc20/data/app/

base.apk!libmonochrome_64.so (Buildld:

#26 pc 00000000030f3f40 /data/app/ac433fc399c6918e19ed2987d803c88499557d3f) ~~JOeVQK2j4wnreygioRHgtA==/com.google.android.trichr

~~JOeUOK2j4wnreygioRHgtA==/com.google.android.trichr omelibrary 683416333-Sq2Qz61RSe5PBlq0F060aQ==/

omelibrary_683416333-Sq2Qz61RSe5PBlq0F06QaQ==/base.apk!libmonochrome_64.so (Buildld:

ac433fc399c6918e19ed2987d803c88499557d3f) #27 pc 00000000030f3a5c /data/app/

~~J0eUQK2J4wnreygioRHgtA==/com.google.android.trichro melibrary_683416333-Sq2Qz61RSe5PBlq0F0600Q==/

ac433fc399c6918e19ed2987d803c88499557d3f) base.apk libmanochrome_64.so (Buildld: #28 pc 00000000030f3984/data/app/

--JOeVQK2j4wnreygioRHgtA==/com.google.android.trichr omelibrary_683416333-Sq2Qz61RSe5PBlq0F06QaQ==/base.apkllibmonochrome_64.50 (Buildld:

ac433fc399c6918e19ed2987d803c88499557d3f #29 pc 0000000003222c08/data/app/

~~JOeUQK2j4wnreygioRHgtA==/com.google.android.trichr omelibrary_683416333-Sq2Qz61RSe5PBlq0F060aQ==/

base.apk!libmonochrome_64.sc (Buildld:

ac433fc399c6918e19ed2987d803c88499557d3f) #30 pc 00000000000f57c8 /apex/

com.android.runtime/lib64/bionic/libc.so pthread_start/void*)+208) (Buildld:

802277585401a6103d67lea1f801ed52)

#31 pc 000000000008f1bc /

apex/com.android.runtime/lib64/bionic/

libc.so_start_thread +68) (Buildld:

802277585401a6103d67lea1f801ed52)

```

Clearing cache do fixes the issue, I will try to test a build with my changes, to see if it fixes it.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/ConversationMessages.svelte:75]: `measureElement` is rarely executed, as `ResizeObserver` updates and caches sizes in sizeMap. As a result, the virtualizer doesn’t need to call measureElement directly in most cases. However, it still acts as a fallback and would be used if an element is rendered for the first time without a cached measurement.

MasterK0927:
I wanted to provide a quick update on the status. I tested this PR on a new build using my smartphones and test devices and confirmed that the issue no longer occurs on them. However, we still need to gather feedback from beta users.

As discussed, we should create a separate build with virtual scrolling specifically for beta users.

Before we go forward with that build, I need to make the scrolling and height adjustments more robust.

MasterK0927:
Now its gtg from my side, just a consideration that we can do more optimization in height calculation, but we can leave it for future updates.

Also I had tested the scrolling behavior on desktop, it works pretty smooth now.

ui/src/lib/types.ts
```diff
@@ -253,3 +253,24 @@ export interface FileExtended {
...
+ * - Adjusting scroll positions smoothly as content loads/changes
+ *
+ * Works with dynamic content where:
+ * - Initial sizes are estimated
+ * - Actual measurements happen asynchronously
+ * - Sizes can change with content updates and resizing
+ *
+ * @property height - Current height in pixels
+ * @property measured - True if height is from actual measurement, false if estimated
+ */
```
mattyg: Please no ai-generated docs. Just do a simple 1 line comment explaining what this is used for. This just adds noise.

ui/src/routes/conversations/[id]/ConversationMessages.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/ConversationMessages.svelte:75]: I don't think so. I put a console log within the function and it never executed, even when the size cache was empty. In the examples on the tanstack/virtual site they call this function explicitly.

Side note: unless you're the one adding the comment, please leave comment threads open and wait for the original commenter mark it resolved. Otherwise I have no way of tracking what has actually been resolved or responsed to.

Comment (MasterK0927) [reply ui/src/routes/conversations/[id]/ConversationMessages.svelte:75]: In TanStack docs, they use the ref callback to call measureElement explicitly. In our case, it's just a fallback, so removing it won't affect functionality. What are your views on it?

Comment (mattyg) [reply ui/src/routes/conversations/[id]/ConversationMessages.svelte:75]: This still needs to be removed, unless I'm wrong above.

ui/src/lib/types.ts
```diff
@@ -253,3 +253,10 @@ export interface FileExtended {
   file?: File;
   status: FileStatus;
 }
+
+// Size of DOM element
+// Used when dynamically measuring DOM element included virtual scrolling list
+export interface SizeInfo {
+  height: number;
+  measured: boolean;
```
mattyg, line 261: It looks like anytime a size is added to the cache, `measured` is always set to true. If so, what's the point of it?

ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -1,13 +1,131 @@
...
+        });
+      }
+    });
+
+    return () => resizeObserver.disconnect();
+  });
+
+  // Scrolls to the bottom of the message list
+  // Aligns to end of list with smooth animation
+  export function scrollToBottom() {
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte:125]: Why is this exported?

mattyg:
Hey @MasterK0927 I'm going to take over the PR from here.

From reviewing the tanstack/virtual examples, I think this implementation is overly complex and should be simplified. We now have scroll handling in the parent page, and in the ConversationMessages component, which needs to be remedied. A few other issues, but mostly I'm convinced this can be simplified a lot.

mattyg:
Superseded by #512

PR #508, opened 2025-02-14
feat(release): add bundling of .aab files for playstore release

MasterK0927:
- Adds bundling of `.aab` files and includes them alongside APKs for Google Play Store releases.

PR #509, opened 2025-02-14
Backport/508

MasterK0927:
Backport #508 

We didn't have the `foreground-service` backported to `develop-0.3`, feature flags like holochain_bundled won't work. 

So, instead of cherry-picking, I manually added the command.

### DIFF
--- a/.github/workflows/release-tauri-app.yaml
+++ b/.github/workflows/release-tauri-app.yaml
@@ -118,15 +118,17 @@ jobs:
 
       - name: Build android AAB
         run: |
+          cargo clean
           nix develop .#androidDev --no-update-lock-file --command bash -c "npm run tauri android build -- --aab --target aarch64 --target i686 --target x86_64"
 
       - name: Build android APKs
         run: |
+          cargo clean
           nix develop .#androidDev --no-update-lock-file --command bash -c "npm run tauri android build -- --apk --split-per-abi --target aarch64 --target i686 --target x86_64"
 
       - uses: AButler/upload-release-assets@v3.0
         with:
-          files: src-tauri/gen/android/app/build/outputs/apk/*/release/app-*
+          files: "src-tauri/gen/android/app/build/outputs/apk/*/release/app-*;src-tauri/gen/android/app/build/outputs/**/*.aab"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           release-id: ${{ needs.create-release.outputs.releaseId }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release-tauri-app.yaml
```diff
@@ -116,16 +116,30 @@ jobs:
...
 
-      - uses: AButler/upload-release-assets@v3.0
+      - name: Upload Android APK  
+        uses: AButler/upload-release-assets@v3.0
         with:
           files: src-tauri/gen/android/app/build/outputs/apk/*/release/app-*
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           release-id: ${{ needs.create-release.outputs.releaseId }}
 
+      - name: Upload Android AAB
```
mattyg: We should be able to upload both the .aab and the .apk within the same step. See the "files" parameter in https://github.com/AButler/upload-release-assets?tab=readme-ov-file#inputs

mattyg (changes requested):
Looks like this isn't a backport. Can we do these changes to `develop` and then cleanly backport them?

Might as well address this at the same time: https://github.com/holochain-apps/volla-messages/issues/510

As well as actually testing that the releases build and upload the `.aab` for a draft release on `develop`

mattyg:
> We didn't have the foreground-service backported to develop-0.3, feature flags like holochain_bundled won't work.
>
> So, instead of cherry-picking, I manually added the command.

I'm don't understand how this is relevant. We're not building with those feature flags on `develop-0.3`. 

To backport, cherry-pick the commit that is displayed on github when the PR branch is merged into `develop`

i.e. this one from the original PR 
image

So the command would be `git cherry-pick -m1 d511e71`. Then resolve the merge conflicts.

MasterK0927:
> > We didn't have the foreground-service backported to develop-0.3, feature flags like holochain_bundled won't work.
> > So, instead of cherry-picking, I manually added the command.
> 
> I'm don't understand how this is relevant. We're not building with those feature flags on `develop-0.3`.
> 
> To backport, cherry-pick the commit that is displayed on github when the PR branch is merged into `develop`
> 
> i.e. this one from the original PR image
> 
> So the command would be `git cherry-pick -m1 d511e71`. Then resolve the merge conflicts.

Initially I did the same thing, but the issue is that in `develop` it uses `{matrix.feature}` which is not defined in `develop-0.3`. So I had to manually change the commands instead of just cherrypicking.

MasterK0927:
Necessary changes will be incorporated in backport PR of #511. So, this PR is no longer required and is being closed.

PR #511, opened 2025-02-18
feat(release): include .aab files in release asset uploads

### LINKED ISSUE #510: Include .aab files in release assets for `develop`
Need to include  .aab files in the release assets along with apks. Specific to `develop`, already added in `develop-0.3` with #509

MasterK0927:
### Summary
Resolves #510

### TODO:
- [x] CHANGELOG updated

### DIFF
--- a/.github/workflows/release-tauri-app.yaml
+++ b/.github/workflows/release-tauri-app.yaml
@@ -159,7 +159,7 @@ jobs:
 
       - uses: AButler/upload-release-assets@v3.0
         with:
-          files: src-tauri/gen/android/app/build/outputs/apk/*/release/app-*
+          files: "src-tauri/gen/android/app/build/outputs/apk/*/release/app-*;src-tauri/gen/android/app/build/outputs/bundle/**/*.aab"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           release-id: ${{ needs.create-release.outputs.releaseId }}
 
--- a/.github/workflows/release-tauri-app.yaml
+++ b/.github/workflows/release-tauri-app.yaml
@@ -149,13 +149,13 @@ jobs:
           cargo clean
           nix develop .#androidDev --no-update-lock-file --command bash -c "npm run tauri android build -- --apk --split-per-abi --target x86_64 --features ${{ matrix.feature }} --config src-tauri/tauri.${{ matrix.feature }}.conf.json"
 
-      - name: Rename android APKs
+      - name: Rename android APKs and ABBs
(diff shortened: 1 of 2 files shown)


mattyg (changes requested):
looks like ci is failing because the .aab files need to be renamed, since we're building two versions for holochain_bundled and holochain_service features.

MasterK0927:
Now its gtg

mattyg (changes requested):
Looks good -- can you add a note to the changelog?

mattyg (approved):

mattyg:
Okay ready for backport. I know we were having issues with building for armv7, and I see its excluded from the aab build here.

I'm assuming google play won't accept the .aab without builds for armv7. If that's the case let me know and I'll make sure to prioritize fixing the build issue in holochain.

MasterK0927:
> Okay ready for backport. I know we were having issues with building for armv7, and I see its excluded from the aab build here.
> 
> I'm assuming google play won't accept the .aab without builds for armv7. If that's the case let me know and I'll make sure to prioritize fixing the build issue in holochain.

Backporting it might not work, as this includes foreground service, which is still not backported. 

For the google play store part, we have release builds from 0.3 right? So ig after backporting to 0.3, we can have a draft release from it, and I can give aab files to Volla team for testing in play store to confirm whether the issue you mentioned arises or not.

mattyg:
>  Backporting it might not work, as this includes foreground service, which is still not backported.

This PR has nothing to do with the foreground service, it just adds a step to CI to build an .aab file.

When you backport a PR, you must make sure to only include the changes needed. So just include that step, excluding any code changes specific to building the 'lite' vs 'rich' releases.

> For the google play store part, we have release builds from 0.3 right? So ig after backporting to 0.3, we can have a draft release from it, and I can give aab files to Volla team for testing in play store to confirm whether the issue you mentioned arises or not.

Yes, releases are from `develop-0.3`

.github/workflows/release-tauri-app.yaml
```diff
@@ -149,17 +149,17 @@ jobs:
           cargo clean
           nix develop .#androidDev --no-update-lock-file --command bash -c "npm run tauri android build -- --apk --split-per-abi --target x86_64 --features ${{ matrix.feature }} --config src-tauri/tauri.${{ matrix.feature }}.conf.json"
 
-      - name: Rename android APKs
+      - name: Rename android APKs and ABBs
```
mattyg, line 152: Typo

Issue #515, opened 2025-02-25
Allow loading previous set of messages while loading current set


mattyg:
In the conversation page, the logic for loading a previous or current set of messages only executes if there is not already a set of messages being loaded.

AC:
- Modify logic of loading a set of messages to allow loading a *previous* batch at the same time as a *current* batch is loading, and vis-versa

PR #519, opened 2025-03-09
feat: remove  bottom loading spinner from message list when scrolling to bottom

### LINKED ISSUE #514: Remove bottom loading indicator on conversation page
AC:
- No longer display the *bottom* loading indicator when on the conversation page. The top loading indicator should not change.

Requested by Dr Wurtzer

MasterK0927:
### Summary
Resolves #514 

### TODO:
- [ ] CHANGELOG updated

### DIFF
--- a/ui/src/routes/conversations/[id]/ConversationMessages.svelte
+++ b/ui/src/routes/conversations/[id]/ConversationMessages.svelte
@@ -182,11 +182,7 @@
               without breaking scrollToBottom. 
             -->
             {#if row.index === messages.length - 1}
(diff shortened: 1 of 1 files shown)


mattyg (changes requested):
- Please note in the changelog
- `loadingBottom` is now no longer used in this component, and thus is not needed as a prop, and should not be passed from the parent
- This issue is also related and a 1-2 line change (https://github.com/holochain-apps/volla-messages/issues/515), probably makes sense to include here.

PR #521, opened 2025-03-14
Remove bottom loading spinner from message list when scrolling and add concurrent message loading

### LINKED ISSUE #514: Remove bottom loading indicator on conversation page

### LINKED ISSUE #515: Allow loading previous set of messages while loading current set

MasterK0927:
### Summary
Resolves #514 
Resolves #515

### TODO:
- [x] CHANGELOG updated

### DIFF
--- a/ui/src/routes/conversations/[id]/ConversationMessages.svelte
+++ b/ui/src/routes/conversations/[id]/ConversationMessages.svelte
@@ -182,11 +182,7 @@
               without breaking scrollToBottom. 
             -->
             {#if row.index === messages.length - 1}
-              <div class="flex h-4 items-center justify-center">
-                {#if loadingBottom}
-                  <SvgIcon icon="spinner" moreClasses="!h-4" />
-                {/if}
-              </div>
+              <div class="flex h-4 items-center justify-center"></div>
             {/if}
           </div>
         </div>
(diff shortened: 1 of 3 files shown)


ui/src/routes/conversations/[id]/ConversationMessages.svelte
```diff
@@ -182,11 +181,7 @@
               without breaking scrollToBottom. 
             -->
             {#if row.index === messages.length - 1}
-              <div class="flex h-4 items-center justify-center">
-                {#if loadingBottom}
-                  <SvgIcon icon="spinner" moreClasses="!h-4" />
-                {/if}
-              </div>
+              <div class="flex h-4 items-center justify-center"></div>
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessages.svelte:184]: Now that the loading indicator is gone, we don't need this, or the parent if statement, or the comment above

mattyg (changes requested):
see comment

ui/src/routes/conversations/[id]/ConversationMessages.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/ConversationMessages.svelte:184]: Apparently removing this breaks the auto scrolling to bottom when a new message is added. I'm going to keep it as a hacky work around.

mattyg:
Reminder to backport this @MasterK0927

PR #533, opened 2025-03-28
fix(singal-handler): use correct app signal data structure supported on holochain version 0.3

### LINKED ISSUE #532: Message deletion not working on `develop-0.3`
In builds from `develop-0.3`, message deletion doesn't work, possibly due to an issue with the app signal data structure's version on develop-0.3.

MasterK0927:
### Summary
Resolves #532 

### TODO:
- [ ] CHANGELOG updated

### DIFF
--- a/ui/src/store/SignalHandler.ts
+++ b/ui/src/store/SignalHandler.ts
@@ -1,4 +1,4 @@
-import { encodeHashToBase64, type Signal, SignalType } from "@holochain/client";
+import { encodeHashToBase64, type AppSignal } from "@holochain/client";
 import { RelayClient } from "$store/RelayClient";
 import { type RelaySignal, type MessageSignal } from "$lib/types";
 import { encodeCellIdToBase64 } from "$lib/utils";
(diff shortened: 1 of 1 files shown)


mattyg:
I did not mean to merge this. It looks like the other PR `fix/p2p-shipyard-upstream-flake` was actually a continuation of this branch branch, but with additional commits that removed this branch's changes. Is that right? 

Was it intentional to undo this PR's changes? If so, did you mean to close this PR?

If you were just trying to ensure a clean git diff, while also basing the other branch off this one, please set the second PR to merge into the first PR, rather than `develop-0.3`.

I will re-review for the issue this aims to address now.

mattyg:
Looks good. It seems that the interface change to the signal handler was not backported in holochain-client-js, so that's why it looks like an incorrect type according to the docs and typescript warnings. I'll take a look at it.

MasterK0927:
> I did not mean to merge this. It looks like the other PR `fix/p2p-shipyard-upstream-flake` was actually a continuation of this branch branch, but with additional commits that removed this branch's changes. Is that right? 
> 
> Was it intentional to undo this PR's changes? If so, did you mean to close this PR?
> 
> If you were just trying to ensure a clean git diff, while also basing the other branch off this one, please set the second PR to merge into the first PR, rather than `develop-0.3`.
> 
> I will re-review for the issue this aims to address now.

It was meant that, PR for this issue will get merged first, and then the upstream flake one. So we have to first merge it and then update the other PR and merge it.

mattyg:
>It was meant that, PR for this issue will get merged first, and then the upstream flake one. So we have to first merge it and then update the other PR and merge it.

In that case, next time please set the other PR to merge into *this* PR, rather than develop-0.3.

PR #543, opened 2025-04-08
Fixing Support line break and Grow text input box as requested in tic…

### LINKED ISSUE #120: Grow text input box if you type in a long message
Show multiple lines

### LINKED ISSUE #141: Support line break in input field

arvibuddy:
Fixing Support line break and Grow text input box as requested in ticket 120 and 141

Resolves #120
Resolves #141

### TODO:
- [ ] CHANGELOG updated

### DIFF
--- a/ui/src/routes/conversations/[id]/ConversationMessageInput.svelte
+++ b/ui/src/routes/conversations/[id]/ConversationMessageInput.svelte
@@ -78,6 +78,19 @@
     text = "";
     files = [];
   }
+  function autoGrow(el) {
+      el.style.height = 'auto';
+      el.style.height = el.scrollHeight + 'px';
+
+      // Optional: clamp height to 3 lines
+      const maxHeight = parseFloat(getComputedStyle(el).lineHeight) * 5;
+      if (el.scrollHeight > maxHeight) {
+        el.style.overflowY = "auto";
+        el.style.height = maxHeight + "px";
+      } else {
+        el.style.overflowY = "hidden";
+      }
+    }
 </script>
 
 <div class="bg-tertiary-500 dark:bg-secondary-500 w-full flex-shrink-0 p-2">
@@ -89,15 +102,17 @@
       </div>
     </label>
 
-    <div class="flex w-full flex-col">
+    <div class="flex w-full flex-col overflow-y-auto leading-6 max-h-[7.5rem]">
       <!-- svelte-ignore a11y-autofocus -->
-      <input
+      <textarea
         autofocus
         type="text"
         bind:this={ref}
         bind:value={text}
+        rows="1"
         class="bg-tertiary-500 w-full border-0 placeholder:text-sm placeholder:text-gray-400 focus:border-gray-500 focus:ring-0"
         placeholder={$t("common.message_placeholder")}
+        on:input={(e) => autoGrow(e.currentTarget)}
         on:keydown={(e) => {
           if (e.key === "Enter" && !e.shiftKey) {
             e.preventDefault();


MasterK0927:
@arvibuddy 

Please mention the tickets it solves, in the below format

Resolves #ticketNumber

arvibuddy:
@MasterK0927 This is done, Updated the commit message and updated tickets numbers

Resolves #120
Resolves #141

ui/src/routes/conversations/[id]/ConversationMessageInput.svelte
```diff
@@ -78,6 +78,19 @@
     text = "";
     files = [];
   }
+  function autoGrow(el) {
+      el.style.height = 'auto';
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessageInput.svelte:82]:
```suggestion
```
You're overwriting this property in the next line.

ui/src/routes/conversations/[id]/ConversationMessageInput.svelte
```diff
@@ -78,6 +78,19 @@
...
+      el.style.height = el.scrollHeight + 'px';
+
+      // Optional: clamp height to 3 lines
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessageInput.svelte:85]:
```suggestion
      // Clamp height to maximum 5 lines
```
Doesn't seem optional as its implemented here. And looks likes its 5 lines not 3.

ui/src/routes/conversations/[id]/ConversationMessageInput.svelte
```diff
@@ -78,6 +78,19 @@
...
+      const maxHeight = parseFloat(getComputedStyle(el).lineHeight) * 5;
+      if (el.scrollHeight > maxHeight) {
+        el.style.overflowY = "auto";
+        el.style.height = maxHeight + "px";
+      } else {
+        el.style.overflowY = "hidden";
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessageInput.svelte:91]: `overflow: auto` shows a scrollbar when the content is overflowing, and hides it when it is not. So we don't need to set this to hidden. We can remove this `else` clause

ui/src/routes/conversations/[id]/ConversationMessageInput.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/ConversationMessageInput.svelte:82]: Ah I see that this makes the textarea shrink again when content is removed. Wondering if there is a clearer way to implement that. Or if not, make a comment to note the reason for it.

ui/src/routes/conversations/[id]/ConversationMessageInput.svelte
```diff
@@ -89,15 +102,17 @@
       </div>
     </label>
 
-    <div class="flex w-full flex-col">
+    <div class="flex w-full flex-col overflow-y-auto leading-6 max-h-[7.5rem]">
```
Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessageInput.svelte:105]:
```suggestion
    <div class="flex w-full flex-col max-h-[7.5rem]">
```
Seems like neither of these rules do anything, unless I'm missing something

Comment (mattyg) [inline ui/src/routes/conversations/[id]/ConversationMessageInput.svelte:105]:
```suggestion
    <div class="flex w-full flex-col overflow-y-auto leading-6">
```

This breaks rendering when you attach several files such that they overflow the max height. I think the max lines specified in the js is all we need.

https://github.com/user-attachments/assets/156daef8-43e4-4a46-a5a3-efbc5fd3f0a6

mattyg (changes requested):
Thanks, I made some comments inline. Also a few more changes to request:
- styling doesn't match dark mode:
image
- link breaks are not rendered in actual message text:

https://github.com/user-attachments/assets/30dca433-c699-43ca-b03a-7b6e253a5080

- After the textarea expands, and thhe message is submitted, the textarea does not shrink back down to original size

https://github.com/user-attachments/assets/5f7a9224-aa2b-4d40-a556-f1904456bf7d

- When you make a PR, update the CHANGELOG file so we can quickly track all changes included in a release

ui/src/routes/conversations/[id]/ConversationMessageInput.svelte
Comment (mattyg) [reply ui/src/routes/conversations/[id]/ConversationMessageInput.svelte:91]: So then we also don't need to set overflowY dynamically in js above. We can just include the `overflow-y-auto` class in the textarea element

mattyg:
Also there is a UX question we should raise to @wurzer:

On desktop, I can create a newline with Shift-Enter. Hitting Enter still sends the message.

On mobile, Shift-Enter does *not* create a newline. Not sure how you would do that.

I'd suggest that we may want to make the following UX changes:
- just hitting Enter create a newline on desktop and mobile
- user needs to actually tap the submit button to send the message

PR #549, opened 2025-04-21
Adds caching of messages to the local storage

MasterK0927:
### Summary
Extends the caching layer to cache conversation message to the local storage.

### TODO:
- [ ] CHANGELOG updated

### DIFF
--- a/ui/src/config.ts
+++ b/ui/src/config.ts
@@ -27,3 +27,8 @@ export const ZOME_NAME = "relay";
 // Polling intervals, used on conversations page
 export const POLLING_INTERVAL_SLOW = 30 * 1000;
 export const POLLING_INTERVAL_FAST = 2 * 1000;
+
+// Cache
+export const CACHE_KEY = "message_store_cache";
+export const CACHE_VERSION = "1";
+export const CACHE_EXPIRY_MS = 24 * 60 * 60 * 1000; // 1 day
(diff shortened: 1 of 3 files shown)


MasterK0927:
Needs to be thoroughly tested in production env.

MasterK0927:
Needs to be thoroughly tested in production env.

mattyg (changes requested):
I think we would want to implement a generic "cached store" so that it can be used in multiple stores if needed, and its implementation details don't add complexity and noise to the ConversationMessageStore. It may make sense to create a `GenericPersistedCacheStore` that uses the `GenericPersistedStore` under-the-hood, and adds the additional functionality for expiring data or forcing the cache to clear. I'm not sure I understand the need for a "cache version".

Then for integrating this into `ConversationMessagesStore` -- I don't think it makes sense to have an *additional* `persisted`, in addition to the existing `messages` `GenericKeyKeyValueStore`. Now we're storing messages in-memory twice, as well as persisting them to local storage. Instead we should only have 1 in memory store that also persists to local storage.

And finally, I don't think the cached data is actually being used in a way that improves the app & conversation initialization performance. In the `initialize` function we're first loading from the cache, then still awaiting the same process to fetch buckets of messages from the holochain conductor. During that process we fetch the list of message hashes, then compare it to the messages we're currently storing, then fetch the messages we're missing. However we're not comparing the hashes to the cached data.

When that's done, we should run this on an app instance with large conversations to confirm it improves the "loading app data" splashscreen performance, before merging.

zippy:
out of date

PR #563, opened 2025-06-13
Fix/lite build packagename

mattyg:
### Summary
Support the 'lite' and 'rich' builds having different android package names.

To accomplish this we no longer check-in the gen/android project directory into git. Instead we have 2 project directories gen/android-holochain_service and gen/android-holochain_bundled.

Before building or running either variant, the project directory must be copied or symlinked to gen/android. Npm commands and ci scripts are modified to do this.

### TODO:
- [x] CHANGELOG updated


zippy (approved):
This is fantastic, thanks @mattyg.

PR #567, opened 2025-07-18
feature: adds in low-level user-network settings and test

zippy:
### Summary

Adds the ability for the end-user to set all the of the network server URLs a the tauri level and creates invokable commands to retrieve default values, and get & set current values.

Notes:
- Default values are compiled into the app, and are used if there is no user saved values.
- Values saved by the user are placed in the appropriate application data folders for each platform.
- You can determine values has ever been set by invoking get_user_network_config and seeing if the value returned is `null`
- Upon saving values the app is restarted so that they will take effect
- This PR does NOT implement 

### TODO:
- [ ] CHANGELOG updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10424,6 +10424,7 @@ dependencies = [
  "holochain_types",
  "lair_keystore",
  "log",
+ "serde",
  "serde_json",
  "tauri",
  "tauri-build",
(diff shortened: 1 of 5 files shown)


zippy:
merged as part of https://github.com/holochain-apps/volla-messages/pull/575

